PR #62, opened 2026-04-19
active session live state

mysoul12138:
What changed:
- `ChatPanel` now uses `chatStore.isSessionLive(sessionId)` for active pinning and live indicator rendering.
- Added `isSessionLive()` to the chat store to combine streaming and resuming state.
- Updated client tests to cover live-vs-selected behavior.

Why:
- Live sessions were not staying pinned, and the live indicator/icon was not showing correctly.

Root cause:
- The UI was relying on session selection state instead of the session’s live/resuming state.

Checks:
- `npm test -- tests/client/chat-panel.test.ts`
- `npm run build`

### DIFF
--- a/packages/client/src/components/hermes/chat/ChatPanel.vue
+++ b/packages/client/src/components/hermes/chat/ChatPanel.vue
@@ -80,6 +80,16 @@ function sourceSortKey(source: string): number {
   return 0
 }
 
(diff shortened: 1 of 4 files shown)


EKKOLearnAI:
已合入 dev 分支并修复选中态高亮问题，后续随 dev 合入 main。

PR #171, opened 2026-04-23
fix: context-length API returns 200K instead of actual model context

### LINKED ISSUE #169: BUG: Context length always returns 200K (DEFAULT_CONTEXT_LENGTH) instead of actual model context
## Bug Description

The `/api/hermes/sessions/context-length` API always returns `200000` (200K) instead of the actual model context length configured in `~/.hermes/config.yaml`. This causes the Web UI context progress bar to display incorrect percentages and fail to warn users when they approach the real context limit.

## Root Cause

The `getModelContextLength()` function in `dist/server/index.js` has a broken fallback chain:

```javascript
function getModelContextLength(profile) {
  const profileDir = getProfileDir2(profile);
  const model = getDefaultModel(profileDir);
  if (!model) return DEFAULT_CONTEXT_LENGTH;      // 200,000
  return lookupContextFromCache(model) || DEFAULT_CONTEXT_LENGTH;  // 200,000
}
```

The function only has two paths:
1. **Cache lookup** via `lookupContextFromCache(model)` → reads from `~/.hermes/models_dev_cache.json`
2. **Hardcoded default** → `DEFAULT_CONTEXT_LENGTH = 200000` (200K)

**Critical gap**: The function **completely ignores** `model.context_length` from `config.yaml`, even though:
- The hermes-agent (Python backend) has a 9-level resolution chain that prioritizes `config.yaml`s

devilardis:
## Summary

Fixes two bugs causing `/api/hermes/sessions/context-length` to always return `DEFAULT_CONTEXT_LENGTH` (200K) instead of the actual model context length. Closes #169

## Bug 1: `getModelContextLength` ignores `config.yaml` context_length

The function only checks `models_dev_cache.json` (which doesn't exist in default installations) and falls back to the hardcoded 200K default, completely ignoring the user's explicit `model.context_length` setting in `config.yaml`.

This is inconsistent with hermes-agent, where `config.yaml`'s `model.context_length` is the highest-priority override (resolution step 0 in a 9-level chain).

## Bug 2: `getDefaultModel` regex fails when `api_key`/`base_url` come before `default`

The regex assumes `default` is the first child key under `model:`:

```javascript
const match = content.match(/^model:\s*\n\s+default:\s*(.+)$/m)
```

This fails when other fields appear before `default` in config.yaml:

```yaml
model:
  api_key: '1'
  base_url: http://192.168.199.66:18080/v1
  default: Qwen3.6-35B-A3B-Claude-4.6-Opus-Reasoning-Distilled.Q4_K_M
  context_length: 64000
```

When `getDefaultModel` returns `null`, `getModelContextLength` short-circuits to `DEFAULT_CONTEXT_LENGTH` (200K) **before even reaching the cache lookup or config fallback**. This means Bug 2 alone is sufficient to cause the 200K issue.

## Changes

1. **`getDefaultModelRobust()`** — extracts the entire `model:` block first, then searches for `default:` within it. Used as primary model name extractor, with the original `getDefaultModel` as fallback.

2. **`getConfigContextLength()`** — reads `model.context_length` from `config.yaml` as a fallback when `models_dev_cache.json` is unavailable or doesn't contain the model.

3. **Updated `getModelContextLength()`** — new resolution order:
   - `models_dev_cache.json` (existing behavior)
   - `config.yaml` `model.context_length` (new — matches hermes-agent priority)
   - `DEFAULT_CONTEXT_LENGTH` (existing fallback)

## Verified Result

```
Before fix: GET /api/hermes/sessions/context-length → {"context_length": 200000}
After fix:  GET /api/hermes/sessions/context-length → {"context_length": 64000}
```

## Files Changed

- `packages/server/src/services/hermes/model-context.ts` (+56/-3)

### DIFF
--- a/packages/client/src/api/hermes/system.ts
+++ b/packages/client/src/api/hermes/system.ts
@@ -45,6 +45,7 @@ export interface CustomProvider {
   base_url: string
   api_key: string
   model: string
+  context_length?: number
   providerKey?: string | null
 }
 
--- a/packages/client/src/components/hermes/models/ProviderFormModal.vue
+++ b/packages/client/src/components/hermes/models/ProviderFormModal.vue
@@ -29,6 +29,7 @@ const formData = ref({
   base_url: '',
   api_key: '',
   model: '',
+  context_length: '',
 })
 
 const modelOptions = ref<Array<{ label: string; value: string }>>([])
@@ -75,7 +76,7 @@ watch(() => formData.value.base_url, (url) => {
 
 watch(providerType, () => {
   modelOptions.value = []
-  formData.value = { name: '', base_url: '', api_key: '', model: '' }
+  formData.value = { name: '', base_url: '', api_key: '', model: '', context_length: '' }
   selectedPreset.value = null
 })
 
@@ -154,11 +155,13 @@ async function handleSave() {
       ? selectedPreset.value
       : null
 
+    const contextLength = formData.value.context_length ? parseInt(formData.value.context_length, 10) : undefined
     await modelsStore.addProvider({
       name: formData.value.name.trim(),
       base_url: formData.value.base_url.trim(),
       api_key: formData.value.api_key.trim(),
       model: formData.value.model,
+      context_length: contextLength,
       providerKey,
     })
     message.success(t('models.providerAdded'))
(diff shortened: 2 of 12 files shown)


EKKOLearnAI:
https://github.com/EKKOLearnAI/hermes-web-ui/pull/177

PR #176, opened 2026-04-24
fix: getApiKey reads config.yaml + correct port/host from config upstream

yuanyang749:
## Summary

Two fixes for external Hermes gateway mode in hermes-web-ui:

### Bug 1 — getApiKey() returns null without .env

`mgr.getApiKey()` only read from `.env` files. External gateway mode has no `.env`, so the proxy always sent no auth token (401/502).

**Fix:** `getApiKey()` now falls back to reading `platforms.api_server.key` from `config.yaml`.

### Bug 2 — Embedded gateway failure leaves stale port/host

When the embedded gateway fails to start, the `gateways` map was populated with `port`/`host` from the failed embedded instance (wrong port) but `url` from `config.upstream` (correct). Other code reading `port`/`host` got wrong values.

**Fix:** Parse `port`/`host` from `config.upstream` URL when registering the fallback upstream.

### Note

The original `authorization` header passthrough approach was reverted — the browser sends WebUI token, not the gateway API key. Correct auth is now injected via the improved `getApiKey()`.

## Files changed

- `gateway-manager.ts` — getApiKey() config.yaml fallback + correct upstream port/host parsing

### DIFF
--- a/packages/server/src/routes/hermes/proxy-handler.ts
+++ b/packages/server/src/routes/hermes/proxy-handler.ts
@@ -72,7 +72,7 @@ function buildProxyHeaders(ctx: Context, upstream: string): Record<string, strin
     const lower = key.toLowerCase()
     if (lower === 'host') {
       headers['host'] = new URL(upstream).host
-    } else if (lower === 'origin' || lower === 'referer' || lower === 'connection') {
+    } else if (lower === 'origin' || lower === 'referer' || lower === 'connection' || lower === 'authorization') {
       continue
     } else {
       const v = Array.isArray(value) ? value[0] : value
--- a/packages/server/src/routes/hermes/proxy-handler.ts
+++ b/packages/server/src/routes/hermes/proxy-handler.ts
@@ -72,7 +72,7 @@ function buildProxyHeaders(ctx: Context, upstream: string): Record<string, strin
     const lower = key.toLowerCase()
     if (lower === 'host') {
       headers['host'] = new URL(upstream).host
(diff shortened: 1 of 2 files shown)


EKKOLearnAI (changes requested):
Thanks for the PR! The direction is right (external gateway mode has issues), but both fixes have problems:

## Bug 1 — Preserving browser `authorization` header is incorrect

The browser sends the **WebUI's own auth token** (`~/.hermes-web-ui/.token`), which is a completely different auth system from the Hermes gateway's API key. Forwarding this to the gateway won't work — the gateway doesn't recognize this token.

The current flow is:
1. Strip `authorization` from browser headers
2. Inject gateway API key via `mgr.getApiKey()`

The problem is `getApiKey()` returns `null` when there's no `.env` with `API_SERVER_KEY`. The fix should be to make `getApiKey()` correctly read the gateway's API key for external gateway mode, not to blindly forward the browser's WebUI token.

## Bug 2 — `port` and `host` are stale when embedded gateway fails

```ts
const { port, host } = this.readProfilePort(name)
// port/host = failed embedded gateway (e.g. 8643)
const configUpstream = config.upstream.replace(/\/$/, '')
// configUpstream = e.g. http://127.0.0.1:8642
this.gateways.set(name, { port, host, url: configUpstream })
```

`port`/`host` come from the failed embedded gateway (e.g. 8643), but `url` points to `config.upstream` (e.g. 8642). These are inconsistent — if any other code reads `port` or `host`, it gets the wrong value. The fallback should either parse `port`/`host` from `configUpstream`, or not store them at all.

PR #351, opened 2026-04-30
fix(chat): isolate concurrent session events and workspace dialog i18n

etjson:
## Problem

Fixes two user-visible bugs reported in the multi-session concurrent scenario,
plus a related crash uncovered while investigating.

### 1. "Set Workspace" dialog title shows raw i18n key

`ChatPanel.vue` opens the workspace dialog with `t('chat.setWorkspaceTitle')`,
but only `setWorkspace` was defined in `en.ts` / `zh.ts` — `setWorkspaceTitle`
was never added. When i18n cannot resolve a key it falls back to echoing the
key itself, so users saw the literal string `chat.setWorkspaceTitle` as the
dialog title (looking like an "unrendered placeholder").

### 2. AI replies leak into the wrong window when two sessions run concurrently

Repro: open sessions A and B, send a message in each so both runs are in
flight. Regardless of which tab is focused, A's streaming tokens and B's
streaming tokens get interleaved into each other's transcripts.

Root cause is in the client `startRunViaSocket()`:

- The whole app shares a single `/chat-run` Socket.IO connection
  (`connectChatRun()` is a singleton).
- Every `startRunViaSocket()` call registers its own listeners
  (`socket.on('message.delta', ...)`, `socket.on('tool.started', ...)`, etc.)
  on the **same** socket.
- The server does fan events out via `session:${session_id}` rooms, but a
  single socket can join multiple rooms, so **both event streams arrive at
  this socket**.
- The client did not filter by `session_id`. Each listener's closure captured
  its own `sid` and wrote the deltas into its own session — cross-contaminating
  both transcripts.

The server already tags every payload with `session_id` in
[`chat-run-socket.ts`](https://github.com/etjson/hermes-web-ui/blob/main/packages/server/src/services/hermes/chat-run-socket.ts#L281-L289),
so the fix is purely a client-side guard that drops events not belonging to
this run.

### 3. Setting a workspace on a never-persisted session crashes

`setWorkspace` controller called `updateSession(id, ...)` directly, but the
local session row is only created when the user sends the first message.
Setting a workspace before sending any message threw a NotFound.

## Fix

| File | Change |
|------|--------|
| `packages/client/src/i18n/locales/en.ts` | Add `setWorkspaceTitle: 'Set Session Workspace'` |
| `packages/client/src/i18n/locales/zh.ts` | Add `setWorkspaceTitle: '设置会话工作区'` |
| `packages/client/src/api/hermes/chat.ts` | Filter by `session_id` inside `handleEvent`; events not addressed to this run are dropped. Untagged events (legacy server / anonymous run) still pass through for backwards compatibility. |
| `packages/server/src/controllers/hermes/sessions.ts` | `setWorkspace`: lazily create the local session row if it does not exist yet, then update. |

Key client-side diff:

```ts
const expectedSid = body.session_id
const handleEvent = (event: RunEvent) => {
  if (closed) return
  if (expectedSid && event.session_id && event.session_id !== expectedSid) {
    return // not for this run — belongs to a concurrent session
  }
  onEvent(event)
  if (event.event === 'run.completed' || event.event === 'run.failed') {
    cleanup()
    onDone()
  }
}


etjson:
@EKKOLearnAI 看下，我测试可以了，顺便修了一个会话数据隔离

PR #475, opened 2026-05-06
fix: correct session reset mode options in Settings UI

GaoZzr:
## Bug Description

The Session Settings dropdown in the WebUI has two bugs that prevent proper session reset mode configuration:

### Bug 1: `hourly` instead of `daily`
The dropdown uses `value: 'hourly'` but the gateway's `SessionResetPolicy` (in `gateway/config.py`) only recognizes `"daily"`, `"idle"`, `"both"`, and `"none"`. Selecting "Scheduled Only" writes `mode: hourly` to config.yaml, which the gateway silently ignores — effectively behaving as `mode: none`.

### Bug 2: Missing `none` option
The CLI setup wizard (`hermes_cli/setup.py`) offers 4 choices including "Never auto-reset (context lives until /reset or context compression)" → `mode: none`. This is also documented in `cli-config.yaml.example` and explicitly handled in `gateway/session.py`:

```python
if policy.mode == "none":
    return None  # No reset
```

However, the WebUI dropdown only has 3 options, making it impossible to select `none` through the UI.

### `SessionSettings.vue`
- Changed `value: 'hourly'` → `value: 'daily'` to match gateway expectations
- Added `{ label: t('settings.session.modeNone'), value: 'none' }` option
- Added `modeDaily` i18n key (renamed from `modeHourly`)

### i18n locale files (all 8 languages)
- Renamed `modeHourly` → `modeDaily` (same translations)
- Added `modeNone` translation for each language:
  - en: "Never (Manual Only)"
  - zh: "永不（仅手动）"
  - de: "Nie (nur manuell)"
  - es: "Nunca (solo manual)"
  - fr: "Jamais (manuel uniquement)"
  - ja: "なし（手動のみ）"
  - ko: "안함 (수동만)"
  - pt: "Nunca (apenas manual)"

## Impact

- Users who previously selected "Scheduled Only" in WebUI were unknowingly running with an unrecognized mode (effectively `none`). After this fix, they'll get the intended `daily` behavior.
- Users who want to disable auto-reset can now do so through the UI instead of manually editing config.yaml.

## Related

- `gateway/config.py`: `SessionResetPolicy` — documents 4 modes: `"daily"`, `"idle"`, `"both"`, `"none"`
- `hermes_cli/setup.py`: CLI wizard offers all 4 modes
- `cli-config.yaml.example`: `mode: both  # "both", "idle", "daily", or "none"`


EKKOLearnAI:
Pulled these commits onto current main and added the missing zh-TW locale follow-up in #800 while preserving GaoZzr's original commit authorship.

PR #532, opened 2026-05-08
fix: recover orphaned ephemeral chat runs

SulthanZahran1:
## Summary

Fixes a WebUI chat persistence failure where a run can finish in Hermes' temporary `eph_*` session but never get copied back into the visible WebUI session.

This PR keeps the existing architecture intact: WebUI still writes normal chat history to its local SQLite DB, while each model execution uses a fresh Hermes ephemeral session internally. The change makes that bridge durable and recoverable.

## User-visible problem

A user can see a streamed answer during the live chat run, but after reload / another device / server restart the conversation is missing from Chat and History.

The transcript may still exist on disk as:

```text
$HERMES_HOME/sessions/session_eph_*.json
```

…but it is not present in WebUI's local `sessions` / `messages` tables, so the UI has nothing visible to sync or display.

## Root cause

`chat-run-socket` creates a fresh Hermes session per WebUI run:

```ts
const hermesSessionId = session_id
  ? `eph_${Date.now().toString(36)}_${Math.random().toString(36).slice(2, 8)}`
  : undefined
```

Before this PR, the only mapping from visible WebUI session -> ephemeral Hermes session was in memory:

```ts
this.hermesSessionIds.set(session_id, hermesSessionId)
```

On normal completion, WebUI calls `syncFromHermes(localSessionId, hermesSessionId)` to copy assistant/tool messages back into the local DB. That path had two brittle assumptions:

1. The process must keep the in-memory mapping until completion.
2. The ephemeral transcript must be readable from Hermes `state.db` immediately.

If either assumption fails, the raw `session_eph_*.json` file can survive while the visible WebUI session remains incomplete or missing assistant/tool messages.

## What changed

- Adds a small durable `chat_run_ephemeral_runs` table to persist the local-session -> ephemeral-session mapping.
- Records the mapping as soon as a run starts.
- Recovers pending/failed ephemeral runs when `/chat-run` initializes after a server restart.
- Makes `syncFromHermes()` retry normal completion up to 10 times, matching the abort path's more robust behavior.
- Adds a raw session-file fallback when Hermes `state.db` has no ephemeral session data yet:

```text
$HERMES_HOME/sessions/session_${hermesSessionId}.json
```

- Marks ephemeral mappings as `synced` only after local DB materialization succeeds.
- Does not enqueue ephemeral deletion when sync fails, so recoverable transcripts are not deleted before being materialized.
- Handles older raw session files that serialized tool results as user messages with `[Tool result:` prefixes.

## Why this approach is safe

- The fallback only runs for internal `eph_*` sessions that WebUI already created for a visible local session.
- Existing state.db sync remains the primary path.
- The raw-file path uses the same local insertion/update logic as the existing sync path where possible.
- The tracking table is append/update-only for active runs and marks completion; it does not change public session IDs or sidebar filtering.
- Failed syncs remain recoverable because the ephemeral delete queue is only populated after successful materialization.

## Verification

- [x] `npm run test -- tests/server/ephemeral-session-recovery.test.ts --run`
- [x] `npm run test -- tests/server/session-sync.test.ts --run`
- [x] `npm run test -- tests/server --run` — 28 files passed, 225 tests passed, 2 skipped
- [x] `npm run build`

## Notes

This is intentionally scoped to the ephemeral run materialization bug. It does not change the separate session-list sync behavior for normal `webui` / `api_server` sessions.

### DIFF
--- a/packages/server/src/db/hermes/schemas.ts
+++ b/packages/server/src/db/hermes/schemas.ts
@@ -188,6 +188,18 @@ export const GC_SESSION_PROFILES_SCHEMA: Record<string, string> = {
   created_at: 'INTEGER NOT NULL',
 }
 
+export const CHAT_RUN_EPHEMERAL_RUNS_TABLE = 'chat_run_ephemeral_runs'
+
+export const CHAT_RUN_EPHEMERAL_RUNS_SCHEMA: Record<string, string> = {
+  local_session_id: 'TEXT PRIMARY KEY',
+  hermes_session_id: 'TEXT NOT NULL UNIQUE',
+  profile_name: "TEXT NOT NULL DEFAULT 'default'",
+  status: "TEXT NOT NULL DEFAULT 'running'",
+  created_at: 'INTEGER NOT NULL',
+  updated_at: 'INTEGER NOT NULL',
+  last_error: 'TEXT',
+}
+
 // ============================================================================
 // Schema Sync Utilities
 // ============================================================================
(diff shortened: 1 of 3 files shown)


EKKOLearnAI:
hermes聊天记录在200多条的时候会强制上下文压缩，并且丢弃正在对话中的消息，整个聊天记录混乱，之所以不用持久化就是这个原因

PR #558, opened 2026-05-09
fix(session): sync all Hermes session sources

ddandyy74:
Closed by author. The original report mixed in environment-specific details. A sanitized report will be filed separately.


ddandyy74:
Closing because the initial report included environment-specific details. I will resubmit a sanitized fix if needed.

PR #624, opened 2026-05-11
fix: Windows gateway lifecycle — needsRunMode, taskkill cleanup, CLI awareness

a2735444:
**Windows gateway lifecycle fixes**

Fixes three issues affecting Windows:

### 1. needsRunMode excludes Windows
detectInitSystem() correctly returns 'windows-service' for Windows, but needsRunMode includes it in the exclude list, causing gateway start (scheduled task) instead of gateway run (detached child process).

### 2. stop() uses POSIX-only process group kill
On Windows, process.kill(-pid, 'SIGTERM') throws. Added taskkill /F /T /PID fallback.

### 3. hermes-cli.ts missing Windows background mode
Added isWindows check alongside isDocker for background gateway run.

Related: PR #623

### DIFF
--- a/packages/server/src/services/hermes/gateway-manager.ts
+++ b/packages/server/src/services/hermes/gateway-manager.ts
@@ -27,10 +27,11 @@
  *
  * 启动模式：
  *   - 正常系统（macOS/Linux）：hermes gateway start/stop（系统服务管理）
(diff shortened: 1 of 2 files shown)


EKKOLearnAI:
我正在兼容windows，马上就来

PR #682, opened 2026-05-13
fix(chat-run): ensure in-memory messages are processed in resumeSession

zhuzhh:
When a session state exists in memory (sessionMap), the resumeSession() function was directly returning state.messages without processing them through handleMessage(). This caused user messages to be missing from the returned messages list because:

1. User messages are added to memory immediately in handleRun()
2. But when client reconnects and session is in memory, raw messages were returned without format conversion
3. Assistant messages from Hermes DB sync had different format

The fix ensures both loaded and in-memory messages go through handleMessage() for consistent format and complete data.

Changes:
- Modified resumeSession() to process in-memory messages via handleMessage()


EKKOLearnAI:
已重构整个run-chat，当前代码不适用

PR #684, opened 2026-05-13
fix: add missing models for OpenCode Go provider

tevenfeng:
## Summary

Add missing model entries to the OpenCode Go provider preset list in both server and client providers.ts.

## Changes

- Added: `kimi-k2.6`, `deepseek-v4-pro`, `deepseek-v4-flash`, `mimo-v2.5-pro`, `mimo-v2.5`, `qwen3.6-plus`, `qwen3.5-plus`

## Files Modified

- `packages/server/src/shared/providers.ts`
- `packages/client/src/shared/providers.ts`


EKKOLearnAI:
已通过 #700 整合合并，关闭此 PR。

PR #754, opened 2026-05-15
fix: prevent yaml corruption when saving model config via web UI

zjulyy:
## Problem

When using the Hermes Desktop Web UI to change model/provider settings, ~/.hermes/config.yaml gets corrupted with garbage characters, making the entire config unparseable.

## Root Causes

Two independent root causes were identified:

### 1. Non-atomic file write (writeConfigYaml)

The function writes directly to the config file via writeFile(cp, ...). If another Hermes process (CLI web_server, TUI gateway) writes to the same file simultaneously, the output is interleaved - producing broken YAML.

Fix: Write to a temp file first, then atomically rename() to the target path.

### 2. Incompatible YAML quoting (writeConfigYaml)

js-yaml was configured with quotingType that forces ALL string values to be double-quoted. Python's PyYAML (used by the Hermes CLI) does not do this. The round-trip between libraries produces inconsistent formatting and can misalign subsequent writes.

Fix: Remove quotingType and let js-yaml choose quoting naturally, matching PyYAML's behavior.

### 3. Model sub-key deletion (setConfigModel)

The function unconditionally does config.model = {} before setting default and provider, which silently deletes base_url, context_length, and other sub-keys that the CLI depends on.

Fix: Preserve existing model sub-keys by using spread before overwriting.

### DIFF
--- a/packages/server/src/controllers/hermes/models.ts
+++ b/packages/server/src/controllers/hermes/models.ts
@@ -424,7 +424,11 @@ export async function setConfigModel(ctx: any) {
   }
   try {
     const config = await readConfigYaml()
-    config.model = {}
+    // Preserve existing model sub-keys (base_url, context_length, api_mode, etc.)
+    // by spreading the existing model dict. This prevents silent deletion of
+    // keys when the CLI (which stores model as a dict) reads back the config.
+    if (typeof config.model !== 'object' || config.model === null) { config.model = {} }
+    config.model = { ...config.model } // defensive copy to avoid mutation surprises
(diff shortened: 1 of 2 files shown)


EKKOLearnAI:
原子 rename 只能避免读到半写入的目标文件，不能解决并发 writer 互相踩，而且你去掉了我的 config.model={}换成 config.model = { ...config.model } 没有考虑到旧配置残留问题

EKKOLearnAI:
如果单纯只是用ai来修改不去理解整个层面，希望不要再提交了

PR #873, opened 2026-05-20
Fix/deep skill scan recursive

gutanulaif:
Problem
    
    The Web UI skill scanner only scanned two directory levels — directly looking for
    <category>/<skill-name>/SKILL.md. However, many skills are organized in a three-level
    structure: <category>/<sub-category>/<skill-name>/SKILL.md (e.g.,
    mlops/evaluation/lm-evaluation-harness/SKILL.md). The second pass over sub-category
    containers (like evaluation/, inference/, training/) was flat and did not recurse
    into deeper directories, causing 12 out of 13 mlops skills to be invisible.
    
    Additionally, the file content endpoint (pA handler) constructed paths as
    skills/<category>/<skill>/<file> — which fails for nested skills
    (e.g., it looked for skills/mlops/lm-evaluation-harness/SKILL.md instead of
    skills/mlops/evaluation/lm-evaluation-harness/SKILL.md).
    
    Changes
    
    - scanSkillsDir: replaced the flat one-level subdir scan with a recursive
      depth-first search for SKILL.md at every directory level under a category.
    - listFiles handler (sA route): replaced hardcoded join(category, skill) path
      with recursive findSkillDir() so nested skill file browsing resolves correctly.
    - readFile handler (pA route): added a fallback findSkillDir() when the direct
      path returns 404, mirroring the listFiles fix.
    
    Verification
    
    - mlops category: 13 skills displayed (was 1) ✅
    - Nested skill content (lm-evaluation-harness, vllm, axolotl): 200 ✅
    - Non-nested skill (huggingface-hub): no regression ✅
    - Reference files in nested skills: accessible ✅
    - All 20 other categories: zero regression ✅


EKKOLearnAI:
https://github.com/EKKOLearnAI/hermes-web-ui/pull/876 做了点小改动

PR #910, opened 2026-05-21
feat(bridge): per-session worker isolation for true concurrent execution

iZhangDM:
## Summary

BridgeBroker now spawns one WorkerProcess per session instead of one per profile. Each session gets its own Python subprocess with isolated global state, matching TUIs natural per-process concurrency model.

### Architecture Change

- **Before:** All sessions within a profile share one WorkerProcess → serialized by AgentPool._run_lock
- **After:** Each session gets its own WorkerProcess → truly concurrent, just like TUI

### Changes

- BridgeBroker._workers: profile → session_id keyed
- Routing maps: _run_session, _approval_session, _compression_session
- WorkerProcess now accepts separate key/profile params
- _worker_for_session creates per-session isolated subprocesses
- _gc_idle_workers cleans up idle session workers

### Files

1 file: packages/server/src/services/hermes/agent-bridge/hermes_bridge.py (+96/-76)

### DIFF
--- a/packages/server/src/services/hermes/agent-bridge/hermes_bridge.py
+++ b/packages/server/src/services/hermes/agent-bridge/hermes_bridge.py
@@ -1457,7 +1457,8 @@ class WorkerProcess:
     STARTUP_TIMEOUT_SECONDS = 120
     REQUEST_TIMEOUT_SECONDS = 120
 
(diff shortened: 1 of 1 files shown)


EKKOLearnAI:
目前先按照profile锁，如果要执行并行任务，请先用不同profile来执行

PR #956, opened 2026-05-23
feat: add deep links for sessions, history, and group chat

werserk:
## Summary
- Adds URL-addressable chat sessions via /#/hermes/session/:sessionId
- Adds History deep links via /#/hermes/history/session/:sessionId
- Adds Group Chat room deep links via /#/hermes/group-chat/room/:roomId
- Makes route ids the tab-local source of truth so multiple tabs can keep different resources open
- Adds copy-link affordances and e2e coverage for reload/multi-tab behavior

## Test Plan
- npm run test -- tests/client/i18n-coverage.test.ts tests/client/chat-store-thinking.test.ts
- npm run test:e2e -- tests/e2e/history-session-deeplink.spec.ts tests/e2e/group-chat-room-deeplink.spec.ts tests/e2e/chat-session-multitab.spec.ts --project=chromium
- npm run test:e2e -- tests/e2e/chat-streaming.spec.ts --project=chromium
- npm run build

Note: LaTeX rendering is intentionally kept in separate PR #954.


EKKOLearnAI:
Thanks for the session deep link work. We pulled it onto the current main branch and made a follow-up adjustment for the new profile access model.

The main reason for the changes is that copied session/history links need to carry the target Hermes profile. Without that, a user can open a valid session link while their browser is currently using a different active profile, and the frontend will load the wrong profile-scoped session list. The result is that the session appears missing even though the user may have access to it.

The follow-up PR makes session and history deep links profile-aware by adding the profile query parameter to copied links, loading deep-linked sessions with that profile, and routing chat socket resume/run calls through the session profile instead of only the current active profile. We also moved group-chat room copy-link and clone-room actions into the room right-click menu as requested.

Follow-up PR: https://github.com/EKKOLearnAI/hermes-web-ui/pull/962

PR #983, opened 2026-05-24
fix: dispatch unknown slash commands to the agent bridge (/plan, /spike, /goal, …)

paulocavallari:
## Summary

The Web UI was eating any `/something` input that wasn't one of the nine hard-coded "session commands" (`/usage`, `/clear`, `/title`, `/abort`, `/queue`, `/compress`, `/steer`, `/destroy`, `/status`). Skill commands like `/plan`, `/spike`, `/tdd`, plus persistent commands like `/goal`, never reached the Python bridge — they were rewritten to `{name: 'status', rawName: 'plan'}` and the gateway answered with `Unknown bridge command: /plan`.

The CLI (`cli.py::process_command`) and gateway adapters (`gateway/run.py::_handle_message`) resolve these the right way, using the canonical helpers in `hermes-agent`. The Web UI just never grew the same hook. This PR adds the missing dispatch.

## Repro (before)

In Web UI:
```
/plan write a function that adds two numbers
```
→ Server logs: `Unknown bridge command: /plan`. Skill never loaded.

## Fix (two layers)

**1. `session-command.ts::parseSessionCommand`** returns `null` for unknown aliases instead of fabricating `{name: 'status', rawName: 'plan'}`. Unknown slash inputs now fall through to `handleBridgeRun` like any normal user message.

```diff
   const rawName = match[1].toLowerCase()
   const name = COMMAND_ALIASES[rawName]
-  if (!name) return { name: 'status', rawName, args: match[2]?.trim() || '' }
+  // Only return a parsed command for KNOWN bridge-level aliases. Unknown
+  // slash inputs (e.g. /plan, /spike, /goal, /tdd) must fall through to
+  // handleBridgeRun so the Python bridge can resolve them as skill
+  // commands or surface them to the agent as regular user input.
+  if (!name) return null
   return { name, rawName, args: match[2]?.trim() || '' }
```

**2. `hermes_bridge.py::_run_chat`** resolves `/skill-name` (and `/bundle-name`) inputs **after** prepersisting the literal slash command to the session DB and **before** calling `AIAgent.run_conversation()`. Resolution uses upstream helpers (`agent.skill_commands.build_skill_invocation_message`, `agent.skill_bundles.build_bundle_invocation_message`) so the Web UI shares the canonical injection logic with the CLI and gateway — no duplicated SKILL.md parsing, no drift if those helpers change.

The Python hook is pure-additive and wrapped in defensive try/except so import failures or missing skills silently fall back to the original message — never breaks a run.

## Behavior after

| Input | Before | After |
|-------|--------|-------|
| `/plan criar plano de aula` | `Unknown bridge command: /plan` | DB sees `/plan ...`; model receives full SKILL.md content + instruction (matches CLI/Telegram/Discord) |
| `/spike` / `/tdd` / `/goal` / any of the 88+ skill commands | Same error | Resolved as the corresponding skill |
| `/foobar` (unknown, not a skill) | Same error | Flows through as a plain user message (matches CLI behavior) |
| `/usage`, `/clear`, `/abort`, `/title`, `/compress`, `/steer`, `/queue`, `/destroy` | Worked | Worked (untouched) |

## Verification

Tested live on a self-hosted Hermes Web UI v0.6.0 deployment with the kiro-go profile (88 skill commands, including `/plan`):

- `/plan verifique se a correção dos comandos está funcionando` → SKILL.md was injected as user message with the standard `[IMPORTANT: The user has invoked the "plan" skill...]` activation header, plus the trailing user instruction. The model executed plan-mode behavior identical to CLI.
- Bridge logs confirmed: `Resolved /plan slash command for session <sid> (2519 chars)`.
- Existing bridge commands (`/usage`, `/clear`) keep working unchanged.

## Risk

Low. The TS change tightens behavior (less false-positive command parsing) without removing any working code path. The Python change is additive — a try/except wrapper around two helper calls — and falls back to the original message on any failure.

## Notes for reviewers

- The Python helpers (`agent.skill_commands`, `agent.skill_bundles`) are part of `hermes-agent` proper and are already imported elsewhere in `hermes_bridge.py`'s runtime path. They're stable upstream APIs (used by `cli.py` and `gateway/run.py`).
- The `_run_chat` hook runs **after** `_prepersist_user_message` so the user's literal `/plan` text is what gets stored in session history (matching CLI behavior where the typed command is preserved in transcripts), while the resolved skill content is what the model sees.
- No new dependencies. No config changes. No schema changes.

### DIFF
--- a/packages/server/src/services/hermes/agent-bridge/hermes_bridge.py
+++ b/packages/server/src/services/hermes/agent-bridge/hermes_bridge.py
@@ -17,8 +17,10 @@ import hashlib
 import importlib.util
 import json
 import locale
+import logging
 import os
 import queue
+import re
 import signal
 import shutil
 import socket
@@ -33,7 +35,10 @@ from contextlib import contextmanager
 from dataclasses import dataclass, field
 from pathlib import Path
 from urllib.parse import urlparse
-from typing import Any, Callable
+from typing import Any, Callable, Optional
+
+
+logger = logging.getLogger("hermes_bridge")
 
 
 DEFAULT_ENDPOINT = "tcp://127.0.0.1:18765" if os.name == "nt" else "ipc:///tmp/hermes-agent-bridge.sock"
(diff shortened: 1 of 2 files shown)


paulocavallari:
Thanks for the review! Pushed `d36d6ff` addressing all five points:

1. ✅ **Hoisted imports** — `agent.skill_bundles` and `agent.skill_commands` resolved once at module load. ImportError now surfaces as a startup `WARNING` instead of silent per-request DEBUG noise. Helpers reused via module-level callables.
2. ✅ **`_SLASH_COMMAND_RE`** compiled once at module scope.
3. ✅ **Single guard** — the redundant `startswith("/")` + length check pair is gone; regex match is the single source of truth.
4. ✅ **Narrowed per-request excepts** to `(AttributeError, KeyError, TypeError, ValueError)` with `exc_info=True` so future bugs surface in DEBUG tracebacks rather than being silently downgraded.
5. ✅ **Outer catch-all removed.**

Verified end-to-end on a live deployment after rebuild — `/plan` still resolves cleanly, no startup WARNINGs about missing skill modules, and the contract (literal `/...` persisted to history, resolved SKILL.md sent to model, graceful fallback on failure) is unchanged.

PR #987, opened 2026-05-24
fix(web): keep session alive across transient socket disconnects (mobile background)

paulocavallari:
## Summary

Mobile users — particularly on iOS Safari and the iOS Hermes Web UI app wrapper — see `Error: Socket disconnected: ping timeout` whenever they background the app for more than ~30s. The error stays sticky even after they reopen the app: the chat session won't update, run streaming is dead, and the only fix is to manually reload the session via the side panel.

The agent run on the server keeps streaming the whole time. The bug is purely client-side: every disconnect (transient or not) is treated as fatal and tears down the session's event handlers permanently.

## Repro

1. Start a long-running agent task (anything that takes 60s+) on Web UI from a phone.
2. Lock the phone or switch to another app for ~30s.
3. Return to Hermes Web UI.

**Before:** "Error: Socket disconnected: ping timeout" pinned to the message. No new tokens or tool events arrive even though the run is still going on the server. Selecting the same session from the sidebar manually does refresh — but only via an explicit `resumeSession()` call, not the running stream.

**After:** the in-progress run keeps streaming as if nothing happened. The user sees the latest tokens / tool events when they come back to the foreground.

## Root cause

In `packages/client/src/api/hermes/chat.ts::startRunViaSocket`, every `disconnect` event (other than `'io client disconnect'`) calls `handleSocketError`, which sets `closed = true`, deletes the session from `sessionEventHandlers`, and surfaces the error to the UI:

```ts
const handleSocketDisconnect = (reason: string) => {
  if (closed || reason === 'io client disconnect') return
  handleSocketError(new Error(`Socket disconnected: ${reason}`))
}
socket.once('disconnect', handleSocketDisconnect)
```

But socket.io is configured with `reconnection: true, reconnectionAttempts: Infinity` — it WILL come back. The client just refuses to recognize it. Once `closed = true`, the per-event guards (`if (closed) return`) drop every subsequent `message.delta`, `tool.completed`, etc. on the floor.

iOS makes this trivially reproducible because the OS suspends WebView JavaScript after ~30s in background, so `ping timeout` happens basically every time the user switches apps.

## Fix

Two-part client-only change in `packages/client/src/api/hermes/chat.ts`:

1. **Whitelist transient disconnect reasons.** `transport close`, `transport error`, `ping timeout`, `parse error` are now treated as recoverable: the handler logs nothing and lets socket.io reconnect on its own.
2. **Re-emit `resume` on reconnect.** When socket.io fires `connect` after a transient disconnect, the client emits `resume` so the server re-joins the session room (`socket.join('session:<sid>')`) and starts forwarding events to this socket again. The server-side `socket.on('resume')` handler already exists for explicit page-reload flows — we just hook into it.

The first-connect-vs-reconnect distinction uses `socket.connected` to skip emitting on the initial connect (when `startRunViaSocket` enters with an already-open socket, which is the typical path).

## Why no server change

The server side is fine. `_run_chat` in `hermes_bridge.py` is a worker thread that streams to whoever's currently in the `session:<sid>` room. The room membership is what gets lost on reconnect — and `socket.on('resume')` rebuilds it. Adding the client emit is the only missing piece.

## Behavior matrix

| Disconnect reason | Before | After |
|---|---|---|
| `ping timeout` (mobile bg) | Sticky error, dead stream | Silent reconnect, stream continues |
| `transport close` (network blip) | Sticky error | Silent reconnect |
| `transport error` (TLS/proxy hiccup) | Sticky error | Silent reconnect |
| `parse error` (rare) | Sticky error | Silent reconnect |
| `io server disconnect` (auth fail, kick) | Surfaced as error | Surfaced as error (unchanged) |
| `io client disconnect` (page nav, deliberate close) | Silent | Silent (unchanged) |

## Risk

Very low. The change exclusively narrows what gets treated as a fatal error — anything that's not in the transient list still surfaces exactly as before. socket.io's reconnect machinery is already configured for `Infinity` attempts, so the only thing missing was the client trusting it.

## Verification

Tested live on:
- iOS Safari → Web UI on remote VM behind nginx WebSocket proxy. Background app for 5 minutes → return → in-progress run still streaming.
- Android Chrome → same setup, same result.
- Brief Wi-Fi-to-cellular handoff → no visible interruption.
- Genuine server kick (kill the gateway PID mid-run) → still surfaces a fatal error in the UI as expected.

## Notes for reviewers

- No new dependencies, no API changes, no config changes.
- Server's `socket.on('resume')` handler is at `packages/server/src/services/hermes/run-chat/index.ts:211` — same handler used by the standalone `resumeSession()` export in chat.ts, so we share the proven path.
- We deliberately don't listen for the `resumed` reply in this code path — its main side effect (server-side `socket.join`) is what we need; the messages array would duplicate what the UI already has from the initial load + delta stream.

### DIFF
--- a/packages/client/src/api/hermes/chat.ts
+++ b/packages/client/src/api/hermes/chat.ts
@@ -583,6 +583,24 @@ export function startRunViaSocket(
 
   let closed = false
   const socket = connectChatRun(body.profile)
+
+  // Disconnect reasons that are TRANSIENT — the socket.io client will
+  // auto-reconnect (reconnection: true, attempts: Infinity). We must NOT
+  // tear down the session in these cases, otherwise mobile background
+  // suspension (iOS suspends the WebView in ~30s) leaves a sticky
+  // "Socket disconnected" error that survives even after the user
+  // reopens the app, because closed=true is permanent and event handlers
+  // get deleted from sessionEventHandlers.
+  //
+  // The agent run keeps streaming on the server side; once the client
+  // reconnects we re-join the session room via 'resume' (see below).
+  const TRANSIENT_DISCONNECT_REASONS = new Set<string>([
+    'transport close',
+    'transport error',
+    'ping timeout',
+    'parse error',
+  ])
+
   const handleSocketError = (err: Error) => {
     if (closed) return
     closed = true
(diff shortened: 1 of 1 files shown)


paulocavallari:
Good catch — that was a real listener leak. Pushed `5535e05` addressing it:

- Forward-declared `removeTerminalSocketListeners` so it's in scope at the top of the function.
- Call it inside `handleSocketError` (before `onError`) so teardown is deterministic regardless of which fatal path triggers.

The happy path (run completes / aborts / fails normally) was already calling cleanup via the `onRunCompleted` / `onRunFailed` / `onAbortCompleted` / `onSessionCommand` handlers — only the `connect_error` / fatal-disconnect path was missing it. With the move to persistent `socket.on(...)` listeners on a module-level singleton socket, that path leaked two closures per `startRunViaSocket` invocation. Now it's bounded.

EKKOLearnAI:
Thanks for tracking down the mobile/background reconnect issue. The root cause makes sense: mobile browsers can suspend JS long enough for Socket.IO to hit `ping timeout`, while the server-side run keeps going.

We are going to handle this in our own follow-up PR: #993. It keeps the same direction, but with a couple of adjustments:

- reconnect emits `resume` and also consumes the `resumed` payload, so the UI can catch up with messages/runtime state produced while the client was disconnected;
- the existing-session early-return path avoids attaching extra persistent reconnect listeners;
- fatal disconnects still clean up listeners deterministically;
- `parse error` is not treated as transient for now, since it may indicate a real protocol/proxy issue that should not be silently hidden;
- added focused client tests for transient reconnect, fatal disconnect cleanup, and duplicate-start listener handling.

Appreciate the report and the repro details. They were useful for narrowing the fix.

PR #1029, opened 2026-05-25
[codex] fix bridge terminal env refresh scope

EKKOLearnAI:
## Summary

Builds on #1016 to refresh profile terminal config in the profile worker process only.

Changes from the original PR:

- keep `_refresh_terminal_env()` at worker startup via `_set_worker_profile_env()`
- remove per-agent and per-run refresh calls that mutate process-global `os.environ`
- preserve inherited or `.env` `TERMINAL_*` values when a terminal key is absent from `config.yaml`
- match existing gateway cwd bridge behavior for placeholder cwd values and `~` expansion
- remove direct dependency on `tools.terminal_tool` private globals

## Why

`os.environ` is process-global. In the broker architecture each profile gets its own worker process, so terminal config should be bridged once when that worker starts. Refreshing it again inside `AgentPool.get_or_create()` or `_run_chat()` reintroduces in-process cross-session/profile leakage risk.

## Validation

- `python3 -m py_compile packages/server/src/services/hermes/agent-bridge/hermes_bridge.py`
- `git diff --check HEAD~1..HEAD`
- targeted local `_refresh_terminal_env()` behavior checks for present/missing terminal config, bad YAML, cwd placeholders, `.env` preservation, and Windows-style cwd strings

Co-authored-by: GoldenFish123321 <golden_fish@foxmail.com>


EKKOLearnAI:
Closing for now while we revisit the env isolation strategy and scope.

PR #1046, opened 2026-05-26
feat: ADR-009 singleton updates and Dev Mode preview provider

werserk:
## Summary
ADR-009 makes the Hermes Web UI behave like a singleton workspace: one canonical shell, one active preview slot, a release-first Updates surface, and a Dev Mode-only branch preview provider.

## What changed
- Canonical routing now keeps the Hermes shell at the stable app surface and preserves the legacy redirects needed during the transition.
- The preview registry/service layer now treats previews as a singleton lifecycle with explicit actions to build, remove, promote, and restore the active slot.
- Added the new Updates view so users can see the current release state and entry points without hunting through Dev Mode.
- Wired singleton actions into the sidebar/settings flows and kept branch preview controls under Dev Mode only.
- Added the server-side preview runtime, registry, and branch-build plumbing needed to serve branch previews from the selected worktree.

## Tests
- pnpm build
- pnpm exec tsc --noEmit -p packages/server/tsconfig.json
- pnpm exec vue-tsc -b
- pnpm exec vitest run tests/server/preview-registry.test.ts tests/server/preview-runtime.test.ts tests/server/preview-controller.test.ts tests/server/preview-routes.test.ts tests/server/dev-mode-branch-builds-service.test.ts tests/server/spa-shell.test.ts
- pnpm exec playwright test tests/e2e/auth.spec.ts tests/e2e/authenticated-shell.spec.ts tests/e2e/provider-models.spec.ts tests/e2e/terminal.spec.ts tests/e2e/native-navigation.spec.ts tests/e2e/chat-session-multitab.spec.ts tests/e2e/history-session-deeplink.spec.ts tests/e2e/group-chat-room-deeplink.spec.ts tests/e2e/chat-streaming.spec.ts

## Limitations
- Dev Mode branch preview stays maintainer-only.
- Previews stay path-based; no subdomain routing.
- Obsidian ADR details stay out of the repo; this PR only carries the product/implementation summary.

## Review focus
Please sanity-check the routing transition, singleton preview lifecycle, and Dev Mode branch preview behavior.


EKKOLearnAI:
Thank you for the work you put into this PR.

I’m sorry, but I can’t merge this as-is. The risk is too high and the change set is too large for the current project direction.

I think there may have been a misunderstanding around the original intent. What I meant was closer to cloning a separate copy of the new code under the user’s `.hermes-web-ui` directory for preview/update purposes, not mixing the entire preview/dev-mode/update/routing logic into the main application flow.

If you need this Dev Mode workflow for your own use case, I’d suggest maintaining it in a fork. I really appreciate your contribution and the effort you spent on this project.

EKKOLearnAI:
Also, version preview itself should not require this many changes.

For what I had in mind, the preview path should be a lightweight mechanism: prepare or clone a separate copy under `.hermes-web-ui`, build it in isolation, and expose that copy as a preview. It should not require a large routing rewrite, singleton registry model, Dev Mode UI, and new update architecture all mixed into the main app.

So the concern is not only implementation size, but also that the scope is much broader than the version-preview feature we need right now.

EKKOLearnAI:
To clarify the intended direction: version preview should be much smaller and more isolated than this PR.

The model I have in mind is:

1. Check whether we already have a separate Hermes Web UI source checkout under the user's `.hermes-web-ui` directory.
2. If not, `git clone` it there.
3. Checkout the requested version/branch/commit.
4. Run `npm install` and `npm run build` inside that isolated checkout.
5. Start it as a separate process on another port, for example:

```bash
PORT=8650 \
HERMES_WEB_UI_HOME=/path/to/.hermes-web-ui/preview-webui \
node dist/server/index.js
```

The stable Web UI should only manage that separate preview instance: start, stop, status, logs, and open preview URL. The preview should use its own `HERMES_WEB_UI_HOME`, its own port, and its own process. If needed, it can also use a separate `HERMES_HOME`.

That approach should only need a small surface area: one backend service, one controller/route group, one client API wrapper, a small UI entry, and i18n/tests. It should not require rewriting the main router, adding a `/preview/` runtime, introducing a singleton registry architecture, or mixing a full Dev Mode/update system into the main app.

So the main issue here is scope: the version preview feature should be an isolated preview runner, not a broad app architecture change.

werserk:
image

Thanks for your reply!

I'd like to clarify something.
This PR is currently in draft status, as it will be further refined and further divided. Essentially, I've only outlined the work being done so that both you and developers interested in contributing are aware of the plans for this development vector.

I'm not sure what impression you've gotten of what's happening now, but I'd like to present the interim results in this area.

This will essentially be an improvement to the update system, as we discussed in the previous PR. I've deemphasized dev mode, as requested, and developed a standard user flow.

Since the project is evolving extremely rapidly, it can break user builds. Therefore, there are two modes.

1. Standard
The user (after all, most of those using this solution are at least a bit of a programmer) can update the build or preview it using the /preview/ path. Afterwards, if everything is ok, it switches the preview to the current build.
2. Dev mode
The user connects their URL/something else as a source for possible updates. This way, they can easily switch between their work and your current, fresh build, and the developers can also quickly test new features.

I could, of course, close both PRs and stop developing this topic, but I think it's extremely important. Please review the attached materials and reconsider. I'm aware of the issue regarding the scope of changes; I'll try to reduce it this week. Again, that's why the PRs are in draft.

I wanna something like this in result. Just dont have enough time to complete it right now. 
image
image
image

If your opinion remains the same and this feature doesn't interest you, let me know. I'll continue working on the fork.

werserk:
And yes, the scope has grown because these features are tightly coupled. Right now, it's much easier for me to refine a full-fledged framework and then continue testing other features through it than to decompose it right now for a neat merge.

PR #1064, opened 2026-05-27
fix(security): add SSRF protection to provider model fetch

WenhuaXia:
## Summary

The `/api/hermes/provider-models` endpoint and `config-helpers.fetchProviderModels()` accept user-controlled `base_url` and fetch it server-side without any IP validation, enabling SSRF attacks.

## What was fixed

- Added `packages/server/src/lib/ssrf-protection.ts` — DNS resolve + private/reserved IP check (IPv4 & IPv6)
- SSRF check in `fetchProviderModelList` controller (POST `/api/hermes/provider-models`)
- SSRF check in `config-helpers.fetchProviderModels()` (used by custom provider discovery)

## Blocked addresses

- `127.0.0.0/8` (loopback)
- `10.0.0.0/8`, `172.16.0.0/12`, `192.168.0.0/16` (RFC 1918)
- `169.254.0.0/16` (link-local / AWS metadata)
- `0.0.0.0`, `::`, `::1`
- `fc00::/7`, `fe80::/10` (IPv6 unique-local / link-local)
- `localhost` hostname

## Validation

- `npx vitest run tests/server/ssrf-protection.test.ts` — 15 tests pass
- Related model controller tests pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/server/src/controllers/hermes/models.ts
+++ b/packages/server/src/controllers/hermes/models.ts
@@ -801,6 +801,13 @@ export async function fetchProviderModelList(ctx: any) {
       return
     }
 
+    const { isSafeUrl } = await import('../../lib/ssrf-protection')
+    if (!(await isSafeUrl(baseUrl))) {
+      ctx.status = 400
+      ctx.body = { error: 'base_url must be a public URL' }
+      return
+    }
+
     const base = baseUrl.replace(/\/+$/, '')
     const modelsUrl = /\/v\d+\/?$/.test(base) ? `${base}/models` : `${base}/v1/models`
     const headers: Record<string, string> = {}
--- a/packages/server/src/lib/ssrf-protection.ts
+++ b/packages/server/src/lib/ssrf-protection.ts
@@ -1,0 +1,55 @@
+import dns from 'dns'
+import net from 'net'
+
+const PRIVATE_V4_PREFIXES = [
+  '10.', '127.', '169.254.',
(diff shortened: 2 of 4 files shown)


EKKOLearnAI:
Thanks for working on the SSRF protection.

I understand the security concern here, but I do not plan to merge this PR in its current form.

Hermes Web UI needs to support local and LAN-hosted model providers, for example Ollama, LM Studio, or OpenAI-compatible services running on `localhost`, `127.0.0.1`, `192.168.x.x`, `10.x.x.x`, etc. Blocking all loopback/private addresses would break a core usage pattern for self-hosted and local deployments.

Also, this endpoint does not request an arbitrary user-provided URL directly. The server normalizes the configured provider base URL and appends `/v1/models` or `/models`, so the SSRF impact is reduced compared with a fully arbitrary URL fetch. That said, this does not eliminate the risk completely: timing, status differences, compatible internal endpoints, redirects, and metadata addresses can still matter.

So I agree the area deserves hardening, but the approach needs to be more compatible with local/LAN model use. A better direction would be something like:

- keep local and LAN provider URLs usable by default
- always block high-risk metadata/link-local targets such as cloud metadata endpoints where appropriate
- disable redirects for model discovery requests
- apply short timeouts and response-size limits
- only parse/return standard model IDs, without exposing raw response bodies or detailed internal errors
- optionally add a stricter deployment mode for public/multi-user servers

For now, I am going to leave this PR unmerged.

PR #1102, opened 2026-05-28
[codex] fix virtual list outline anchor jumps

EKKOLearnAI:
## Summary

- Make virtual list anchor navigation render the target message range before aligning the requested outline anchor.
- Align anchors using real DOM geometry over several animation frames instead of relying on one estimated offset plus `scrollIntoView`.
- Add a regression test covering navigation to an anchor that starts outside the current virtual range.

## Root Cause

Conversation outline jumps can target messages that have never been rendered by the virtual list. The previous implementation scrolled once using estimated item heights, then tried to find the anchor DOM node on the next tick. For long jumps, accumulated height estimation error could keep the target message outside the rendered range, so the anchor lookup and final alignment were unreliable until the user clicked near the area again.

## Validation

- `npx vue-tsc -b`
- `npm test -- tests/client/virtual-message-list-anchor.test.ts`
- `npm test -- tests/client/virtual-message-list-anchor.test.ts tests/client/tool-trace-visibility.test.ts tests/client/kanban-task-drawer.test.ts`

### DIFF
--- a/packages/client/src/components/hermes/chat/VirtualMessageList.vue
+++ b/packages/client/src/components/hermes/chat/VirtualMessageList.vue
@@ -36,11 +36,15 @@ const scrollerRef = ref<HTMLElement | null>(null);
 const scrollTop = ref(0);
 const viewportHeight = ref(0);
 const heightVersion = ref(0);
+const forcedRange = ref<{ start: number; end: number } | null>(null);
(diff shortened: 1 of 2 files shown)


EKKOLearnAI:
Closing per request; abandoning this outline anchor scroll approach.

PR #1169, opened 2026-05-31
fix(skills): include symlinked skill directories

### LINKED ISSUE #1027: Skills installed as symlinks not shown in WebUI (scanSkillsDir ignores isSymbolicLink)
## Summary

Skills installed as symlinks (e.g. `agent-browser`) are silently filtered out by `scanSkillsDir()` and never appear in the WebUI skills management page.

## Root Cause

In `packages/server/src/controllers/hermes/skills.ts`, the filter at line 205 uses:

```ts
.filter(e => e.isDirectory() && !e.name.startsWith('.'))
```

Node.js `readdir({ withFileTypes: true })` returns `Dirent` objects. `Dirent.isDirectory()` does **not** follow symlinks — it returns `false` for a symlink pointing to a directory. Therefore any skill installed as a symlink (e.g. via `hermes` CLI or manually) is excluded.

## Fix

Add `e.isSymbolicLink()` to the filter:

```diff
- .filter(e => e.isDirectory() && !e.name.startsWith('.'))
+ .filter(e => !e.name.startsWith('.') && (e.isDirectory() || e.isSymbolicLink()))
```

## Reproduce

1. Create a symlink skill: `ln -s /path/to/skill ~/.hermes/skills/my-skill`
2. Open WebUI → Skills page
3. `my-skill` is missing

## Environment

- hermes-web-ui v0.6.1
- Node.js >=23.0.0
- Linux (symlinks common when sharing skills between profiles or pointing to a shared workspace)

qiontoo:
## Summary
- include symlinked directories when scanning skills
- reuse the same visible-skill-entry predicate for root, category, and recursive scans
- add a regression test for a symlinked skill directory

Closes #1027

## Test Plan
- `npx vitest run tests/server/skills-controller.test.ts --reporter=dot`
- `npx vue-tsc -b`
- `npx tsc --noEmit -p packages/server/tsconfig.json`

Note: `npm run build` reached the Vite `rendering chunks...` stage but exceeded the 600s tool timeout in this environment; focused server/client type checks and the affected test pass.

### DIFF
--- a/packages/server/src/controllers/hermes/skills.ts
+++ b/packages/server/src/controllers/hermes/skills.ts
@@ -2,6 +2,7 @@ import { mkdir, readdir, readFile, stat, writeFile } from 'fs/promises'
 import { homedir } from 'os'
 import { join, resolve } from 'path'
 import { createHash } from 'crypto'
+import type { Dirent } from 'fs'
 import {
   readConfigYamlForProfile, updateConfigYamlForProfile,
   safeReadFile, extractDescription, listFilesRecursive,
@@ -138,7 +139,7 @@ function readUsageStats(usageContent: string | null): Map<string, UsageStats> {
 }
 
 async function findSkillDirByName(rootDir: string, skillName: string): Promise<string | null> {
-  let entries: import('fs').Dirent[]
+  let entries: Dirent[]
   try {
     entries = await readdir(rootDir, { withFileTypes: true })
   } catch {
@@ -188,6 +189,10 @@ async function resolveSkillDirFromConfig(
   return null
 }
 
+function isVisibleSkillEntry(entry: Dirent): boolean {
+  return !entry.name.startsWith('.') && (entry.isDirectory() || entry.isSymbolicLink())
+}
+
 /**
  * Scan for skills at different directory depths.
  *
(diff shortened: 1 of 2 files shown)


hanzckernel:
Thanks for keeping this focused on #1027. I checked the current symlink-skill fix against latest `main`; this PR looks like the right one to recycle rather than opening another duplicate PR.

One hardening gap before merge: the current diff treats symlinked entries as visible directories, but it does not resolve real paths or guard against cyclic links. A safer shape would:

- follow symlinked skill directories only after `stat()` confirms the target is a directory;
- use `realpath()` / visited-set protection while recursively collecting nested skill/category entries;
- keep hidden-entry filtering for symlinked entries too;
- cover both flat symlinked skills and symlinked entries under a category, including a cycle/hidden-link regression.

I have a local branch with that narrower shape if you want me to adapt it into this PR or open a replacement, but my preference is to recycle this focused PR if you are OK refreshing it.

hanzckernel:
I opened #1609 as a refreshed replacement for this focused #1027 fix.

Credit to this PR and @qiontoo for the original symlink-skill direction. The replacement keeps the same scope, but rebases the fix onto latest `main` and adds extra hardening for realpath/cyclic symlink handling plus flat/category regression coverage.

Link: #1609

hanzckernel:
Closing this PR as superseded/stale: #1027 has already been fixed and closed by the merged replacement PR #1609, and this branch now conflicts with latest `main`.

Thanks for the original focused symlink-skill fix direction. If there is still a symlink scanning edge case that #1609 does not cover, please feel free to open a fresh, narrower PR against latest `main` and call out the remaining delta from #1609.

PR #1228, opened 2026-06-01
i18n: add russian translation

vanja-san:
New file


EKKOLearnAI:
Thanks for adding the Russian locale. Could you please wire it into the i18n registry as well? Right now `ru.ts` is added, but the app will not load or show it because `ru` is not registered.

Please update:

- `packages/client/src/i18n/messages.ts`: import `./locales/ru`, add `ru` to `supportedLocales`, and include it in `rawMessages`.
- `packages/client/src/components/layout/LanguageSwitch.vue`: add a Russian option, e.g. `{ label: 'Русский', value: 'ru' }`.

After that, the fallback merge should cover any missing keys from English, but without these registrations the new locale file is currently unused.

vanja-san:
@EKKOLearnAI Done!

PR #1249, opened 2026-06-02
[codex] Require explicit runtime release metadata

EKKOLearnAI:
## Summary
- Removes app-version and latest-release fallbacks from desktop runtime manifest resolution.
- Requires the packaged runtime-release.json metadata, or an explicit runtime override, before resolving runtime assets.
- Prevents GitHub runtime downloads from incorrectly probing releases/latest for runtime manifests.

## Why
Runtime assets live under the hermes-0.15.2-runtime release tag, not the app latest release. Falling back to app version/latest can produce URLs like /releases/latest/download/hermes-runtime-win-x64.json, which are invalid for runtime packages.

## Validation
- npm --prefix packages/desktop run build


EKKOLearnAI:
Closing per request; the observed GitHub runtime download failure appears to be a client network issue rather than a code issue.

PR #1265, opened 2026-06-03
Feat/user avatar upload + group-chat avatar sync

SpikeJulia:
Summary:
    Adds user avatar support with full group-chat integration.
    What's included
    User avatar (Settings → Current Account)
    - Upload PNG/JPEG/WebP, auto-compressed to 500KB
    - Generate random avatars via multiavatar
    - Reset to default
    - GET/PUT /api/auth/avatar endpoints

    Group-chat avatar sync
    - Avatars show in chat messages, header stack, and member popover
    - Resolved by numeric userId, not display name
    - No re-join required

    Testing
    - 31 new unit tests, 907 total, zero failures
    - TypeScript: zero errors

    17 files, +997 -29

PR #1273, opened 2026-06-03
Feat: user avatar upload and group-chat avatar sync

EKKOLearnAI:
## Summary
- bring over the user avatar upload and group-chat avatar sync work from #1265
- persist the authenticated user id for group-chat members so display names can differ from login usernames
- prevent group-chat member avatar fallback from assigning another user's avatar

## Validation
- npm run test -- tests/server/group-chat-avatar.test.ts tests/server/user-avatar.test.ts tests/server/group-chat-member-sync.test.ts tests/client/group-chat-store-streaming.test.ts tests/client/auth-avatar.test.ts
- npm run build

Based on #1265.


SpikeJulia:
Glad to see the avatar feature merged! Thanks @EKKOLearnAI  for polishing it with i18n, tests, and fixing that edge case regarding authUserId for group members (good catch on the display name inconsistency!).
Learned a lot from your follow-up commits on handling UUID+timestamp seeds for random avatars. Looking forward to contributing more features to the project!

PR #1298, opened 2026-06-03
fix chat bridge resume after web restart

ystwjove:
## Summary
  - keep the agent bridge alive across Web UI shutdowns by default
  - reattach to existing bridge runs when a chat session resumes after restart
  - add bridge manager and resume polling coverage

  ## Tests
  - npx tsc -p packages/server/tsconfig.json --noEmit
  - npx vitest run tests/server/run-chat-bridge-resume.test.ts tests/server/agent-bridge-manager.test.ts
  - real 8648 login smoke test with admin / 123456


EKKOLearnAI:
shudown 需要杀掉bridge的

Issue #1300, opened 2026-06-03
Add opt-out flags for gateway autostart and bundled skill injection


poisdahl:
## Summary

In Docker/self-hosted deployments where `hermes-web-ui` is mounted against an existing Hermes Agent home, the Web UI currently performs two startup-side effects by default:

1. It checks/starts profile gateways from `packages/server/src/index.ts` via `ensureProfileGatewaysRunning()`.
2. It injects bundled Web UI skills into target Hermes skill directories via `HermesSkillInjector().injectMissingSkills()`.

This is convenient for all-in-one installs, but risky for dashboard-only/admin deployments where another service already owns the Hermes gateway lifecycle and skills are managed separately.

## Why this matters

- Web UI startup can compete with an already-running production gateway.
- In Docker, managed gateway behavior is effectively auto-enabled through `shouldUseManagedGatewayRunForAutostart()`.
- The skill injector can overwrite same-named target skill directories because `injectIntoTarget()` removes the target skill dir before copying bundled content.
- Operators do not currently have a clear env/config way to run Web UI as a dashboard/admin surface while preserving existing gateway lifecycle and curated skills.

## Proposed behavior

Add explicit env opt-outs, defaulting to the current behavior:

- `HERMES_WEB_UI_DISABLE_GATEWAY_AUTOSTART=1`
- `HERMES_WEB_UI_DISABLE_SKILL_INJECTION=1`

When set:

- skip both `ensureProfileGatewaysRunning()` calls in `startRuntimeServicesBeforeListen()` / `startRuntimeServicesAfterListen()`
- skip startup bundled skill injection in `bootstrap()`
- log a clear bootstrap message so operators can see why startup behavior changed

A minimal patch shape is:

```ts
function envFlagEnabled(name: string): boolean {
  const value = String(process.env[name] || '').trim().toLowerCase()
  return ['1', 'true', 'yes', 'on'].includes(value)
}

function gatewayAutostartDisabled(): boolean {
  return envFlagEnabled('HERMES_WEB_UI_DISABLE_GATEWAY_AUTOSTART')
}

function skillInjectionDisabled(): boolean {
  return envFlagEnabled('HERMES_WEB_UI_DISABLE_SKILL_INJECTION')
}
```

Then gate the two startup gateway checks and the startup skill injection block in `packages/server/src/index.ts`.

## Optional follow-ups

[...]

PR #1301, opened 2026-06-03
feat: allow disabling gateway autostart and skill injection

poisdahl:
## Summary

Adds two explicit startup opt-out flags for dashboard-only/self-hosted deployments:

- `HERMES_WEB_UI_DISABLE_GATEWAY_AUTOSTART=1`
- `HERMES_WEB_UI_DISABLE_SKILL_INJECTION=1`

When set, the server skips gateway autostart checks and bundled skill injection during bootstrap while preserving current default behavior when the flags are unset.

This addresses #1300.

## Why

Some deployments mount `hermes-web-ui` against existing Hermes Agent homes where gateway lifecycle and skills are managed outside the Web UI. In those setups, startup-side effects can compete with production gateways or overwrite curated bundled-skill directories.

## Validation

- `git diff --check`
- `npm ci --ignore-scripts`
- `npm run build`
- Verified with Homebrew `node v26.0.0` and `npm 11.12.1` on macOS
- Verified against current upstream `main` at `1d19be2aeaba466c7b348260d4b1acc416fc6f9d`
- Local deployment validation: these flags keep `hermes-web-ui` healthy in dashboard-only Docker deployments next to existing Hermes Agent containers/gateways, without mutating gateway lifecycle or runtime skills at startup.

Note: a separate verification attempt on a host with Node `v18.19.1` failed before evaluating this patch because the project requires Node `>=23.0.0` and Vite requires Node `20.19+` / `22.12+`.

Disclosure: this PR was prepared with AI assistance (OpenAI Codex).

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -86,13 +86,30 @@ function isDesktopRuntime(): boolean {
   return String(process.env.HERMES_DESKTOP || '').trim().toLowerCase() === 'true'
 }
 
+function envFlagEnabled(name: string): boolean {
+  const value = String(process.env[name] || '').trim().toLowerCase()
+  return ['1', 'true', 'yes', 'on'].includes(value)
+}
+
+function gatewayAutostartDisabled(): boolean {
+  return envFlagEnabled('HERMES_WEB_UI_DISABLE_GATEWAY_AUTOSTART')
+}
+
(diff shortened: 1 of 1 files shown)


EKKOLearnAI:
https://github.com/EKKOLearnAI/hermes-web-ui/pull/1307

PR #1315, opened 2026-06-04
fix: keep agent bridge alive on restart but stop it on shutdown

ystwjove:
## Summary

  - resume active bridge runs after the Web UI process restarts
  - keep the agent bridge alive for intentional Web UI restarts
  - shut down the attached agent bridge on real Web UI stop/shutdown
  - update the CLI restart/stop signal handling so restart uses SIGUSR2 and stop uses SIGTERM

  ## Tests

  - npx vitest run tests/server/agent-bridge-manager.test.ts tests/server/shutdown.test.ts tests/server/
  run-chat-bridge-resume.test.ts
  - npx tsc -p packages/server/tsconfig.json --noEmit

### DIFF
--- a/bin/hermes-web-ui.mjs
+++ b/bin/hermes-web-ui.mjs
@@ -422,6 +422,10 @@ function startDaemon(port) {
 
     fetch(healthUrl).then(res => {
       if (res.ok) {
+        const listeningPid = recoverPidFromPort()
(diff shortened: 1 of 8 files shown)


EKKOLearnAI:
https://github.com/EKKOLearnAI/hermes-web-ui/pull/1320

PR #1333, opened 2026-06-04
fix(chat): merge reasoning across tool cycles into single assistant message

SpikeJulia:
Problem：
When tool.started fires within a run, activeAssistantMessageId was cleared, causing subsequent reasoning.delta events to create a new separate assistant message just for the thinking block. This produced one "思考过程" bubble per tool cycle instead of one per run.
Root Cause
    1. tool.started unconditionally cleared activeAssistantMessageId
    2. reasoning.delta required isStreaming=true — but tool.started turned it off
Fix
    - tool.started: stop clearing activeAssistantMessageId
    - reasoning.delta: skip isStreaming check, resume streaming if paused
    Applied to both live-stream and history-replay handlers.
    Verification
    - ✅ 152 test files / 949 tests pass
    - ✅ vue-tsc compiles clean
    - ✅ harness:check passes
    - ✅ Production build succeeds
    - ✅ Manually tested — multi-tool agent reasoning now merges correctly


EKKOLearnAI:
https://github.com/EKKOLearnAI/hermes-web-ui/pull/1343

PR #1370, opened 2026-06-06
fix: use page reload on logout to clear Pinia store data

hanlehui:
## Summary
- Replace `router.replace` with `window.location.href = ""` in logout handler
- Force full page reload to clear all Pinia store state
- Prevent stale data when switching between accounts

## Problem
When users logout and login with a different account, Pinia store data from the previous session persists in memory. This can cause minor issues with stale state.

## Solution
Use `window.location.href = ""` instead of `router.replace({ name: "login" })` to trigger a full page reload. This ensures all Pinia stores are completely reset when switching accounts.

## Impact
- Users will experience a brief page reload on logout
- Clean state guaranteed for each new login session
- No more stale data issues when switching accounts

## Validation
- [x] Tested logout flow
- [x] Verified page reload clears all store state
- [x] Confirmed login works correctly after reload

### DIFF
--- a/packages/client/src/api/client.ts
+++ b/packages/client/src/api/client.ts
@@ -124,7 +124,7 @@ export async function request<T>(path: string, options: RequestInit = {}): Promi
     clearApiKey()
     emitAuthNotice('expired')
     if (router.currentRoute.value.name !== 'login') {
-      router.replace({ name: 'login' })
+      window.location.href = ''
     }
     throw new Error('Unauthorized')
   }
@@ -136,7 +136,7 @@ export async function request<T>(path: string, options: RequestInit = {}): Promi
         clearApiKey()
         emitAuthNotice('expired')
         if (router.currentRoute.value.name !== 'login') {
-          router.replace({ name: 'login' })
+          window.location.href = ''
         }
       } else {
         emitAuthNotice('forbidden')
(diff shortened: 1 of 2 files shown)


hanlehui:
## 💡 Future Optimization Suggestion

This PR fixes the immediate issue by using `window.location.href` to force page reload on logout. However, a more elegant long-term solution would be:

### Recommended Approach: Reset stores on login

Instead of modifying 3 logout paths (manual logout, 401, 403), reset all Pinia stores when login succeeds:

```typescript
// utils/store-reset.ts
export function resetAllStores() {
  useAppStore().$reset()
  useChatStore().$reset()
  useSettingsStore().$reset()
  // ... other stores
}

// LoginView.vue - after successful login
async function handlePasswordLogin() {
  const sessionToken = await loginWithPassword(username.value.trim(), password.value)
  setApiKey(sessionToken)
  resetAllStores()  // ← Add this
  router.replace("/hermes/chat")
}
```

### Benefits:
- Single point of change (1 location vs 3)
- No page reload = better UX (no white flash)
- More maintainable and explicit

### Note:
Current stores use Composition API (setup stores) which don't have built-in `$reset()`. Would need to either:
1. Switch to Options API style, or
2. Add manual `reset()` methods to each store

This PR is a pragmatic fix; the store reset approach is the ideal long-term solution.

EKKOLearnAI:
https://github.com/EKKOLearnAI/hermes-web-ui/pull/1383

PR #1387, opened 2026-06-07
    fix: add DEFAULT values to session_usage schema to prevent NOT NULL constraint errors

dreamor:
Fix: NOT NULL constraint failed: session_usage.updated_at

    Problem

    When visiting the Hermes Web UI, users may encounter:

    > Error: NOT NULL constraint failed: session_usage.updated_at

    The session_usage table has updated_at defined as INTEGER NOT NULL without a default value. When an INSERT doesn't provide this column, SQLite throws a constraint error.

    Root Cause

    The SCHEMA in usage-store.ts only defines 4 columns but the database has 10 columns. The ensureTable() migration drops/adds columns on every startup, causing transient inconsistencies.

    Fix

    1. Added DEFAULT 0 to updated_at — inserts without it won't fail
    2. Synced SCHEMA to include all 10 columns present in the database
    3. Updated updateUsage() to INSERT all columns matching actual DB structure

    Changes

    - packages/server/src/db/hermes/usage-store.ts: Schema and INSERT fixes


EKKOLearnAI:
https://github.com/EKKOLearnAI/hermes-web-ui/pull/1401

PR #1438, opened 2026-06-09
fix(server): include file paths in API content blocks

qdivan:
## Summary
- Include the local file path when API file content blocks are converted to text input.
- Keep image content block behavior unchanged so API image blocks still send image data without adding local image paths.

## Validation
- `npm run test -- tests/server/run-chat-content-blocks.test.ts`
- `npm run build`


EKKOLearnAI:
api-run 现在已经弃用了

PR #1512, opened 2026-06-12
feat: add integrated terminal panel to ChatView

Xul54299-ux:
Adds integrated terminal panel using xterm.js + node-pty.


EKKOLearnAI:
暂时不考虑这么做

PR #1571, opened 2026-06-15
fix(config): read v12+ providers dict alongside legacy custom_providers list

### LINKED ISSUE #1570: [Bug]: Studio ignores v12+ `providers:` dict schema — only reads legacy `custom_providers:` list
### Hermes Web UI Version

v0.6.15

### Hermes Agent Version

main (post-v12 schema; affects all v12+ users)

### Bug Description

Hermes Studio's model/provider discovery only reads `config.yaml`'s **legacy** `custom_providers:` *list* and does not understand the **v12+ `providers:` *dict*** schema that Hermes Agent itself emits and considers canonical.

As a result, any user whose `~/.hermes/config.yaml` was migrated by Hermes (v11→v12) — or who wrote it in the modern dict shape — sees their custom providers **completely missing** from Studio's "Models / Providers" page, even though `hermes` CLI, the dashboard, and the agent loop resolve them fine.

### Steps to Reproduce

1. Take any custom provider configured the modern way:

   ```yaml
   model:
     default: minimax-m3
     provider: volcengine-coding

   providers:
     volcengine-coding:
       api: ${ARK_CODING_BASE_URL}
       key_env: ARK_CODING_API_KEY
       default_model: minimax-m3
       models:
         - minimax-m3
         - kimi-k2.6
         - glm-5.1
         - deepseek-v4-pro
         - deepseek-v4-flash
       discover_models: false

   custom_providers: []
   ```

2. Run Hermes Studio against this profile.
3. Observe Studio's model picker / `Settings → Models` page: `volcengine-coding` is missing entirely. Models that Studio *does* show come from `auth.json` credential-pool entries (e.g. a separate `deepseek` credential), not from `config.yaml`.
4. `hermes model` CLI, the agent loop, and the messaging gateway all resolve `volcengine-coding` correctly with the same config.

### Expected Behavior

Studio should treat `providers:` (dict) and `custom_providers:` (list) as equivalent input shapes, the same way Hermes Agent itself does. Custom providers declared in the v12+ dict shape should appear in the Studio UI with their models, and the `provider-model-catalog.json` cache should refresh entries for them.

### Actual Behavior

Studio reads `custom_providers:` only. v12+ dict-shaped providers are silently dropped.

### Root cause

Hermes Agent ran into this exact problem in its own runtime back in April and fixed it in commit [`c449cd1af`](https://github.com/NousResearch/hermes-agent/commit/c449cd1af58c00225df29364a2c67c203d8b4582) — *"fix(config): restore custom providers after v11→v12 migration"*. The fix introduces:

- `hermes_cli/config.py::providers_dict_to_custom_providers()` — translates a `providers:` dict entry into the legacy list-shaped record (with `_normalize_custom_provider_entry()` doing field aliasing: `apiKey`/`api_key`, `key_env`/`api_key_env`, `default_model`/`model`, etc.).
- `hermes_cli/config.py::get_compatible_custom_providers()` — returns a deduplicated unified view across **both** schemas. The commit message is explicit:

  > all runtime resolvers read from `custom_providers` — after migration, named custom endpoints silently stop resolving […]. Add `get_compatible_custom_providers()` that reads from both config schemas (legacy `custom_providers` list + v12+ `providers` dict), normalizes entries, deduplicates, and returns a unified list. Update ALL consumers […].

Studio is the last consumer that has not been updated. In the current `dist/server/index.js` (v0.6.15), the two relevant sites both read only the legacy list:

**Catalog refresh** (`l5I` in the minified output, called from startup `[model-catalog-cache] refreshing N configured provider catalogs`):

```js
let d = Array.isArray(e.custom_providers) ? e.custom_providers : [];
for (let n of d) { … }
```

**Model resolution per profile** (`mP`, the function that builds the `/api/hermes/available-models` response — i.e. what the browser renders):

```js
let Y = Array.isArray(c.custom_providers) ? c.custom_providers : [];
let u = await Promise.allSettled(Y.map(async o => {
  if (!o.base_url) return null;
  …
  let w = [...new Set([o.model, ...i].filter(Boolean))];
  return { providerKey: h, label: o.name, base_url: p, models: w, api_key: o.api_key || "", builtin: Pv(h) };
}));
```

Neither site falls back to `providers:`. There is **no** `providers_dict_to_custom_providers` / `getCompatibleCustomProviders` helper anywhere in the bundled server (`grep` returns zero matches).

### Suggested fix

Mirror Hermes Agent's compatibility layer in Studio's TypeScript so both schemas resolve identically:

```ts
// pseudo, mirrors hermes_cli/config.py:get_compatible_custom_providers
function getCompatibleCustomProviders(cfg: any): NormalizedCustomProvider[] {
  const out: NormalizedCustomProvider[] = [];
  const seenKey = new Set<string>();
  const seenTriple = new Set<string>();

  // 1) v12+ providers: dict
  if (cfg?.providers && typeof cfg.providers === "object" && !Array.isArray(cfg.providers)) {
    for (const [providerKey, entry] of Object.entries(cfg.providers)) {
      const norm = normalizeCustomProviderEntry(entry, providerKey);
      if (norm) appendUnique(norm, providerKey, out, seenKey, seenTriple);
    }
  }

  // 2) legacy custom_providers: list
  if (Array.isArray(cfg?.custom_providers)) {
    for (const entry of cfg.custom_providers) {
      const norm = normalizeCustomProviderEntry(entry, "");
      if (norm) appendUnique(norm, "", out, seenKey, seenTriple);
    }
  }

  return out;
}
```

`normalizeCustomProviderEntry()` should accept the same field aliases Hermes does — at minimum: `api`/`url`/`base_url`, `apiKey`/`api_key`, `key_env`/`api_key_env`/`apiKeyEnv`, `model`/`default_model`/`defaultModel`, `models` as either a list of strings or a dict, `transport`/`api_mode`, `discover_models`, `extra_body` — and resolve `${ENV_VAR}` references against the profile's `.env` so the catalog refresher can authenticate the live `/v1/models` probe.

Then both call sites change their `Array.isArray(cfg.custom_providers) ? … : []` line to `getCompatibleCustomProviders(cfg)`. The dict→list translator already exists upstream as `providers_dict_to_custom_providers` if a port is preferred over a re-implementation: <https://github.com/NousResearch/hermes-agent/blob/main/hermes_cli/config.py#L3942>.

### Workarounds for users hitting this today

Two options that don't require waiting on a fix:

1. **Migrate the entry from `providers:` (dict) to `custom_providers:` (list).** Hermes still understands the legacy shape (that's what the compatibility layer is for), so it keeps working. Downside: contradicts what `hermes config migrate` produces, and the next `hermes` CLI write may convert it back.

2. **Add the provider through Studio's own UI** ("Add Custom Provider" → fill in name/base_url/api_key/model). Studio writes it back to `custom_providers:` (list), so it then sees its own entry. Downside: duplicates an entry that already exists in `providers:`, and `hermes config` warnings about duplicate provider names may appear.

Both are workarounds, not solutions; the right fix is the compatibility layer above.

### Logs / Error Messages

No errors in `~/.hermes-web-ui/logs/server.log`. The catalog refresh logs `[model-catalog-cache] refreshing 1 configured provider catalogs` — the "1" is the unrelated `deepseek` credential from `auth.json`. The dict-shaped `volcengine-coding` is simply not seen.

`~/.hermes-web-ui/cache/provider-model-catalog.json` after refresh:

```json
{
  "version": 1,
  "providers": {
    "deepseek|https://api.deepseek.com|all": {
      "provider": "deepseek",
      "models": ["deepseek-v4-flash", "deepseek-v4-pro"],
      "source": "live",
      "profiles": ["default"]
    }
  }
}
```

(The `volcengine-coding` provider with its 5 models is absent.)

### Environment

macOS 26.5.1, Node v26.3.0, npm 9.6.7, Hermes Studio v0.6.15 (npm), Hermes Agent main checkout at `~/.hermes/hermes-agent`.

### Node Version

v26.3.0

### Additional Context

Related upstream commit (the fix Studio still needs to mirror): NousResearch/hermes-agent@`c449cd1af58c00225df29364a2c67c203d8b4582`.
Related upstream call sites that already use the helper: `hermes_cli/runtime_provider.py`, `hermes_cli/auth_commands.py`, `hermes_cli/main.py`, `agent/auxiliary_client.py`, `agent/credential_pool.py` — i.e. every consumer in core. Studio is the lone holdout.

I'm happy to PR the TypeScript port of the compatibility layer if a maintainer would prefer that over a fresh implementation; let me know which call sites you'd like routed through the new helper.

simfeng:
## Summary

- Add `getCompatibleCustomProviders()` (TS port of Hermes Agent's [`get_compatible_custom_providers()`](https://github.com/NousResearch/hermes-agent/commit/c449cd1af58c00225df29364a2c67c203d8b4582)) that reads provider entries from **both** the legacy `custom_providers:` list and the v12+ `providers:` dict, mirrors Agent's field-alias normalization (`apiKey`/`api_key`, `key_env`/`api_key_env`, `model`/`default_model`, list-or-dict `models`, `transport`/`api_mode`), and deduplicates by `provider_key` and `(name, base_url, model)` triple.
- Route Studio's **read** paths through it — model resolution (`controllers/hermes/models.ts`), catalog refresher (`services/hermes/model-catalog-cache.ts`), context-length lookup (`services/hermes/model-context.ts`), coding-agent resolution (`services/coding-agents.ts`), fun-codex media provider lookup (`controllers/hermes/media.ts`), and the `buildModelGroups()` fallback (`services/config-helpers.ts`).
- Write paths in `controllers/hermes/providers.ts` (POST/PUT/DELETE) intentionally **stay on the legacy list** — that's Studio's existing UI write surface and changing it is out of scope for this bug.
- 21 new Vitest cases for the helper, including the exact `volcengine-coding` config from the issue.

## Why

Hermes Agent's v12 migration converts `custom_providers:` (list) into `providers:` (dict) and deletes the list. Agent itself shipped `get_compatible_custom_providers()` so every runtime consumer in core sees a unified list view across both schemas. Studio was the lone holdout — it only read `Array.isArray(config.custom_providers) ? … : []`, which means **any provider declared in the v12+ dict shape is invisible to Studio's model picker, catalog, and context lookup**, even though `hermes` CLI, the dashboard, and the agent loop resolve them fine.

Symptom: `chat.<domain>` Models page shows nothing for a custom provider whose only definition lives under `providers:` (e.g. `volcengine-coding` with a coding-plan model whitelist that doesn't appear in the upstream `/v1/models` listing). The user has to either workaround it by migrating their config back to the legacy list shape, or duplicate the entry through Studio's UI.

## User / developer impact

- **Users on v12+ Hermes configs** (the default for any new install or `hermes config migrate` user) now see their custom providers in Studio's model picker, catalog cache (`provider-model-catalog.json` now includes them), and chat-default resolution.
- **Existing legacy `custom_providers:` users**: zero behavior change — list entries are still read first, dedup keeps the same identity rules (matches Agent's traversal order).
- **Studio's UI write flow**: unchanged. The "Add Custom Provider" form still writes to `custom_providers:`. Hand-editing both schemas is supported (legacy entries take precedence for matching identity triples), matching Agent.

## Validation

```bash
npm run test            # 1579 passed, 2 skipped (across 209 files)
npm run build           # client + server build clean
npm run harness:check   # passes
npx tsc --noEmit -p packages/server/tsconfig.json   # clean
```

New tests in `tests/server/custom-providers-compat.test.ts` (21 cases) cover:

- input validation (non-objects, missing URL, missing name with/without `providerKey`)
- URL aliases (`base_url` / `url` / `api`, first valid wins)
- camelCase aliases, `api_key_env` snake-case alias, `transport`/`api_mode`
- `models` as a list (converted to dict), as a dict (preserved)
- `discover_models` / `context_length` / `rate_limit_delay` / `extra_body` round-trip
- both schemas read individually and merged
- dedup by provider_key and by (name, base_url, model) triple including trailing-slash normalization
- malformed `custom_providers` → empty (matches Agent's bail-out semantics)
- the **exact `volcengine-coding` config** from issue #1570 — Studio now sees the provider with all 5 plan-aliased model names

## Known limitations

- Studio's "Add Custom Provider" UI still writes to `custom_providers:` (list). Migrating the write surface to also recognize/produce the v12 dict is a larger UX change and intentionally out of scope here. Hand-editing the dict in `config.yaml` works correctly today after this fix; the UI just doesn't author the dict shape itself yet.
- The `providers.ts` `attachToCredentialPool` flow likewise still operates on the legacy list. Same reasoning — credential-pool attachment via dict-shaped providers is a separate change.

Closes #1570
Refs NousResearch/hermes-agent@`c449cd1af58c00225df29364a2c67c203d8b4582`

### DIFF
--- a/packages/server/src/controllers/hermes/media.ts
+++ b/packages/server/src/controllers/hermes/media.ts
@@ -4,6 +4,7 @@ import { dirname, extname, isAbsolute, join, resolve } from 'path'
 import { getActiveProfileName, getProfileDir, listProfileNamesFromDisk } from '../../services/hermes/hermes-profile'
 import { config } from '../../config'
 import { readConfigYamlForProfile } from '../../services/config-helpers'
+import { getCompatibleCustomProviders } from '../../services/hermes/custom-providers-compat'
 
 const XAI_VIDEO_GENERATIONS_URL = 'https://api.x.ai/v1/videos/generations'
 const XAI_VIDEO_STATUS_URL = 'https://api.x.ai/v1/videos'
@@ -81,9 +82,7 @@ function buildApiUrl(baseUrl: string, pathWithV1: string): string {
 
 async function resolveFunCodexProvider(profile: string): Promise<FunCodexProvider | null> {
   const hermesConfig = await readConfigYamlForProfile(profile)
-  const customProviders = Array.isArray(hermesConfig.custom_providers)
-    ? hermesConfig.custom_providers as any[]
-    : []
+  const customProviders = getCompatibleCustomProviders(hermesConfig)
   const provider = customProviders.find(entry => String(entry?.name || '').trim() === APIKEY_IMAGE_PROVIDER)
   const apiKey = String(provider?.api_key || '').trim()
   const baseUrl = String(provider?.base_url || '').trim()
--- a/packages/server/src/controllers/hermes/models.ts
+++ b/packages/server/src/controllers/hermes/models.ts
@@ -3,6 +3,7 @@ import { existsSync, readFileSync } from 'fs'
 import { join } from 'path'
 import { getActiveEnvPath, getActiveAuthPath, getActiveProfileName, getProfileDir, listProfileNamesFromDisk } from '../../services/hermes/hermes-profile'
 import { readConfigYaml, readConfigYamlForProfile, updateConfigYaml, updateConfigYamlForProfile, fetchProviderModels, buildModelGroups, PROVIDER_ENV_MAP } from '../../services/config-helpers'
+import { getCompatibleCustomProviders } from '../../services/hermes/custom-providers-compat'
 import { buildProviderModelMap, PROVIDER_PRESETS } from '../../shared/providers'
 import { getCopilotModelsDetailed, resolveCopilotOAuthToken, type CopilotModelMeta } from '../../services/hermes/copilot-models'
 import { readAppConfig, writeAppConfig, type ModelVisibilityRule } from '../../services/app-config'
@@ -317,9 +318,9 @@ async function buildAvailableForProfile(
     currentDefault = String(modelSection.default || '').trim()
     currentDefaultProvider = String(modelSection.provider || '').trim()
     if (currentDefaultProvider === 'custom' && currentDefault) {
-      const cps = Array.isArray(config.custom_providers) ? config.custom_providers as any[] : []
+      const cps = getCompatibleCustomProviders(config)
       const match = cps.find(
-        (cp: any) => cp.base_url?.replace(/\/+$/, '') === String(modelSection.base_url || '').replace(/\/+$/, '')
+        (cp) => cp.base_url?.replace(/\/+$/, '') === String(modelSection.base_url || '').replace(/\/+$/, '')
           && cp.model === currentDefault,
       )
       if (match) currentDefaultProvider = providerKeyForCustom(String(match.name || ''))
@@ -391,9 +392,7 @@ async function buildAvailableForProfile(
     }
   }
 
-  const customProviders = Array.isArray(config.custom_providers)
-    ? config.custom_providers as Array<{ name: string; base_url: string; model: string; api_key?: string }>
-    : []
+  const customProviders = getCompatibleCustomProviders(config)
   const customFetches = await Promise.allSettled(
     customProviders.map(async cp => {
       if (!cp.base_url) return null
@@ -404,7 +403,11 @@ async function buildAvailableForProfile(
       const builtinCatalogModels = isBuiltinProviderKey(providerKey)
         ? PROVIDER_MODEL_CATALOG[builtinProviderKey] || builtinPreset?.models || []
         : []
-      let models = [...new Set([cp.model, ...builtinCatalogModels].filter(Boolean))]
+      // Pull configured models from the v12+ providers.<name>.models dict (if any)
+      // alongside the legacy single `model` field; both contribute to what the
+      // UI shows. `models` is normalized to a dict shape regardless of source.
+      const configuredModels = cp.models ? Object.keys(cp.models) : []
+      let models = [...new Set([cp.model, ...configuredModels, ...builtinCatalogModels].filter(Boolean) as string[])]
       const cachedModels = getCachedProviderModels(modelCatalogCache, providerKey, baseUrl)
       if (cachedModels) models = [...new Set([...models, ...cachedModels])]
       return { providerKey, label: cp.name, base_url: baseUrl, models, api_key: cp.api_key || '', builtin: isBuiltinProviderKey(providerKey) }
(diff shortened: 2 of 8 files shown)


EKKOLearnAI:
https://github.com/EKKOLearnAI/hermes-studio/pull/1581

Issue #1094, opened 2026-05-28
[Feature]: 建议添加环境变量支持配置登录后生成token的有效期


Penghui-liu:
### Problem Statement

建议添加环境变量支持配置登录后生成token的有效期

### Proposed Solution

添加环境变量

### Priority

Critical - blocking my usage

### Willing to Contribute?

- [ ] Yes, I'd like to submit a PR
- [ ] Yes, but I need guidance
- [ ] No, I don't have time

PR #1637, opened 2026-06-17
feat: configure user JWT expiry

qiontoo:
## Summary\n- Add AUTH_JWT_EXPIRES_SECONDS to configure password-login user JWT lifetime.\n- Keep the existing 30-day default when the variable is absent or invalid.\n- Add regression coverage for the configured expiry.\n\nCloses #1094\n\n## Validation\n- `npm run test -- tests/server/user-auth.test.ts -t "uses AUTH_JWT_EXPIRES_SECONDS for issued login JWTs"`\n- `npm run test -- tests/server/user-auth.test.ts`\n- `npx tsc --noEmit -p packages/server/tsconfig.json`\n- `npm run harness:check`\n- `npm run build`


EKKOLearnAI:
Thanks for adding the configurable login JWT expiry. The direction looks good, but please tighten up two things before merge:

1. `AUTH_JWT_EXPIRES_SECONDS` should be parsed strictly. The current `parseInt` accepts partial values such as `7200abc` or `30 days` as valid numbers, while the PR says invalid values fall back to the 30-day default. Please only accept a pure positive integer string and add regression coverage for invalid values falling back to the default.

2. Please document the new environment variable as part of this PR, including that the unit is seconds and the default is 30 days. At minimum this should be added to the server auth/config env documentation where the other env vars are listed.

PR #1638, opened 2026-06-17
fix: include symlink folders in workspace picker

qiontoo:
## Summary\n- Include non-hidden symlinked directories in the workspace folder picker.\n- Add regression coverage using a symlinked workspace directory.\n- Add the required chat-chain change fragment.\n\nCloses #1456\n\n## Validation\n- `npm run test -- tests/server/sessions-controller.test.ts -t "lists workspace folders that are directories or symlinked directories"`\n- `npm run test -- tests/server/sessions-controller.test.ts`\n- `npx tsc --noEmit -p packages/server/tsconfig.json`\n- `npm run harness:check`\n- `npm run build`

### DIFF
--- a/docs/chat-chain-changes/2026-06-18-workspace-symlink-folders.md
+++ b/docs/chat-chain-changes/2026-06-18-workspace-symlink-folders.md
@@ -1,6 +1,6 @@
 ---
 date: 2026-06-18
-commit: 55a2294e
+commit: fc1df8d8
 feature: Workspace folder picker symlink support
 impact: Non-hidden symbolic-link directories are now returned by `/api/hermes/workspace/folders`, so Windows junctions and directory symlinks show up in the workspace selector instead of being filtered out.
 ---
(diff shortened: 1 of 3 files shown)


EKKOLearnAI:
Thanks for fixing the workspace picker issue. The direction is right, but please tighten two things before merge:

1. Please verify symlink targets are actually directories before returning them. The current filter accepts every symlink:

```ts
.filter(e => (e.isDirectory() || e.isSymbolicLink()) && !e.name.startsWith('.'))
```

That can include symlinks to files or broken symlinks, which would show up as folders and then fail when the user tries to open them. Please use `stat(join(fullPath, e.name))` or equivalent for symlinks and include only targets where `isDirectory()` is true. A regression test for a symlink-to-file or broken symlink would be useful.

2. Please fix the chat-chain fragment commit value. `docs/chat-chain-changes/2026-06-18-workspace-symlink-folders.md` currently says `commit: 55a2294e`, which appears to belong to another PR. This PR head is `fc1df8d8`.

PR #1681, opened 2026-06-19
fix(goal): increase goal evaluate timeout from 5s to 120s

WenhuaXia:
## Summary
- Bump \`BRIDGE_GOAL_EVALUATE_TIMEOUT_MS\` from 5s to 120s
- 5s timeout introduced in #1580 is too aggressive for users with remote/slow LLM services
- Goal judge calls silently fail with timeout, causing goal continuation to never enqueue
- Aligns with \`auxiliary.goal_judge.timeout: 120\` default in config.yaml

## Reproduction
1. Set up a remote LLM with 15s+ response time
2. Start a goal in Web UI
3. After first turn, judge times out at 5s → continuation never queued → goal stuck

## Validation
- \`npm run test\` (23 pre-existing failures, no new ones)
- \`npm run build\` passes

## Impact
Fixes goal auto-continuation for users whose LLM response time exceeds 5s.


WenhuaXia:
Closing: PR contains unrelated local commits. Will re-open clean PR.

PR #1734, opened 2026-06-22
fix: persist model selection per chat session

IEatCodeDaily:
## Summary
- route model picker changes through the active chat session when one exists instead of overwriting the global default model
- keep the global model selection only for the no-session/new-chat draft state
- send the session model/provider on every non-coding-agent run so bridge/runtime state cannot drift after reloads or mid-session model changes
- stop treating the loaded model catalog as authoritative for explicit session selections; missing/stale catalog entries no longer reset chats back to the profile default
- show each session's model in the left chat list, using configured model aliases where available
- add a model-switch loading state: disables repeated clicks, keeps the picker open, shows spinner/status feedback, and reports failures inline
- add chat-chain docs fragments and focused store/sidebar/model-selector/server resolver coverage

## Validation
- `npm run harness:check`
- `env NODE_OPTIONS='--max-old-space-size=2048' npm run test -- --pool=forks --poolOptions.forks.singleFork=true --no-file-parallelism tests/server/run-chat-model-config.test.ts tests/client/chat-store-session-command.test.ts tests/client/model-selector-loading.test.ts`
- `nice -n 10 env NODE_OPTIONS='--max-old-space-size=3072' npm run build`

Authored-by: Zephyr (AI Assistant) <raaisalpwardana+zephyr@gmail.com>

### DIFF
--- a/docs/chat-chain-changes/2026-06-22-session-model-persistence.md
+++ b/docs/chat-chain-changes/2026-06-22-session-model-persistence.md
@@ -1,0 +1,6 @@
+---
+date: 2026-06-22
+pr: pending
+feature: Session-scoped model selection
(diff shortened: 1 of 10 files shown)


EKKOLearnAI:
image
我不建议在这里显示模型
全局的切换可以修改我同意

PR #1931, opened 2026-07-04
fix: preserve Studio custom provider headers across runtime selection

nimengbo:
## Summary

This PR now covers two confirmed Hermes Studio paths that prevented `model.default_headers` from reaching WAF-sensitive custom OpenAI-compatible endpoints:

1. Desktop could keep using a stale cached Hermes runtime that did not contain the Agent-side header handling.
2. Even on a current runtime, Studio treated a custom-provider selector such as `custom:liuzheng` as different from the normalized runtime provider `custom`, performed a redundant `AIAgent.switch_model()`, and rebuilt the OpenAI client without the configured headers.

The second path was reproduced on the currently installed release and traced at the final OpenAI client-construction boundary.

## Current Environment

The confirmed reproduction and live verification used:

```text
Hermes Studio: 0.6.29
Hermes Web UI: 0.6.29
Hermes Runtime: 0.18.2
Platform: macOS arm64
Runtime path: ~/.hermes-web-ui/desktop-runtime/hermes/0.18.2/mac-arm64
```

The active runtime metadata came from:

```text
~/.hermes-web-ui/desktop-runtime/active-version.json
```

## Root Cause

Studio sends the selected provider alias:

```text
custom:liuzheng
```

Hermes Agent resolves that selector to the runtime provider:

```text
custom
```

The bridge previously stored only the normalized value in `session.config.provider`. On the next request, this comparison became true:

```text
requested provider: custom:liuzheng
stored provider:    custom
```

Studio therefore considered an unchanged runtime to be a model/provider change and called `AIAgent.switch_model()`.

On Hermes Runtime 0.18.2, the affected switch path reconstructs `_client_kwargs`. Diagnostic capture at the final client factory showed:

```text
agent_init
  default_headers={"User-Agent":"curl/8.7.1"}

switch_model
  default_headers={}

chat_completion_stream_request
  default_headers={}
```

The request then used the OpenAI Python SDK fingerprint and the upstream WAF returned:

```text
HTTP 403: Your request was blocked.
```

A direct request with the same URL, API key, model, and body succeeds when the effective header is:

```text
User-Agent: curl/8.7.1
```

## Changes

- Keep the requested provider selector (for example `custom:liuzheng`) in Studio session metadata instead of replacing it with only normalized `custom`.
- Compare the resolved runtime model, provider, base URL, API mode, and active API credential before invoking `switch_model()`.
- Skip the client rebuild when the effective runtime is unchanged.
- Preserve the existing real model-switch behavior when any resolved runtime field or API credential differs, so credential rotation is not skipped.
- Add regression coverage for a custom provider alias resolving to runtime provider `custom`.
- Retain the earlier stale cached-runtime refresh fix so Desktop upgrades to the packaged Hermes runtime when required.

## Verification

Passed locally:

```text
python3 -m py_compile \
  packages/server/src/services/hermes/agent-bridge/python/bridge_pool.py
```

```text
NODE_ENV=development npx vitest run \
  tests/server/agent-bridge-python-concurrency.test.ts \
  tests/server/agent-bridge-profile-env.test.ts
```

Result:

```text
Test Files  2 passed
Tests       36 passed
```

```text
npm run harness:check
```

```text
Harness check passed
```

```text
NODE_ENV=development npm run build
```

```text
Build completed successfully
```

```text
git diff --check
```

A real Hermes Studio request through the installed Studio Bridge also completed successfully:

```text
status: completed
output: STUDIO_OK
Hermes Studio: 0.6.29
Hermes Runtime: 0.18.2
```

## Notes

`model.default_headers` remains owned by Hermes Agent runtime. This change does not monkey patch the runtime client factory. It fixes Studio's provider-alias/session comparison so Studio no longer performs a destructive no-op client rebuild.


EKKOLearnAI:
我查了 PyPI 的 `hermes-agent==0.18.0` 源码，Agent runtime 侧已经处理了 `model.default_headers`：

- `AIAgent._apply_user_default_headers()` 会读取并合并 `model.default_headers`
- `agent.auxiliary_client._apply_user_default_headers()` 也覆盖了 auxiliary / async client 路径
- shared / per-request OpenAI client 最终使用的是已经合并过的 client kwargs

所以这里不建议在 Studio bridge 里 monkey patch `agent_runtime_helpers.create_openai_client`。`model.default_headers` 是 Hermes Agent runtime 的配置语义，应该由 runtime 自己处理；如果 Studio bundled runtime 落后，正确方向应该是更新/确认 bundled runtime 到包含该修复的版本，而不是在 Studio 里对上游内部 helper 打补丁。

现在仓库里也能看到 runtime packaging 已经指向 Hermes Agent `0.18.0`，而 0.18.0 本身看起来已经没有这个问题。如果仍然有失败路径，麻烦补充具体是哪一个 bundled runtime 版本/路径，以及哪个 client creation path 绕过了 Agent 0.18.0 的 default header merge。

nimengbo:
@EKKOLearnAI Thanks — confirmed. I removed the Studio bridge monkey patch and changed the PR direction.

The failing local path was not Agent 0.18.0 bypassing `model.default_headers`. Studio was still using a cached desktop runtime:

```text
~/.hermes-web-ui/desktop-runtime/hermes/0.16.0/mac-arm64
```

Its installed metadata is:

```text
Name: hermes-agent
Version: 0.16.0
```

That runtime lacks:

- `AIAgent._apply_user_default_headers`
- `agent.auxiliary_client._apply_user_default_headers`

while the repo packaged runtime metadata already points to `hermes-agent 0.18.0`.

So the concrete failing path was:

```text
Studio Desktop → active-version.json → cached runtime 0.16.0 → no Agent 0.18 default_headers merge
```

I updated the PR to rely on the runtime-owned behavior instead:

- removed the bridge-level `create_openai_client` monkey patch
- removed the related bridge patch regression test
- made Desktop bootstrap treat an older cached runtime as stale via `cachedRuntimeNeedsPackagedReleaseUpdate()`
- added runtime-manager coverage for stale cached runtime detection

Fresh verification passed for the targeted checks and build:

```text
npm run harness:check
NODE_ENV=development npx vitest run tests/desktop/runtime-manager.test.ts tests/desktop/runtime-manager-version.test.ts tests/server/agent-bridge-profile-env.test.ts
NODE_ENV=development npm run build
git diff --check
```

Full `npm run test` still has one unrelated existing TTS Blob realm assertion failure in `tests/client/use-speech-tts.test.ts`.

nimengbo:
Closing this PR to replace it with a clean PR focused on the confirmed Hermes Studio 0.6.29 / Runtime 0.18.2 provider-alias client rebuild issue.

PR #1946, opened 2026-07-05
fix: goal kickoff runs losing conversation history due to excludeLastUser truncation

LiangYang666:
## Problem

When the user sends `/goal <prompt>` in Hermes Studio, the agent receives
an empty history (`historyMessages: 0` in bridge.log) and has no context
of the previous conversation.

For example, after the user asks "Explain the difference between forward
proxy and reverse proxy" and the agent replies with a detailed comparison,
the user follows up with `/goal compare the differences`. Instead of
referencing the existing explanation, the agent tries to diff two
nonexistent files, completely unaware of what "compare" refers to.

## Root Cause

The goal flow sets `displayInput: null` on the QueuedRun so the goal text
is not re-displayed as a duplicate user message. This means no `role=user`
message is persisted to the DB for the goal kickoff.

`buildCompressedHistory` → `buildDbHistory({ excludeLastUser: true })`
then looks for the **last `role=user` message** in the DB to exclude it.
Since the goal text was never stored as a user message, it finds the
**previous round's** user message and truncates everything from there —
including the assistant's reply that defines the context for the goal.

Normal chat runs are unaffected because the user message IS persisted
immediately before the run, so `excludeLastUser` correctly excludes the
correct (just-persisted) message.

## Fix

Three changes across four files:

| File | Change |
|------|--------|
| `session-command.ts` | Goal QueuedRun sets `source: 'goal_kickoff'` instead of `'cli'` |
| `handle-bridge-run.ts` | Detect `goal_kickoff` source, pass `{ excludeLastUser: false }` to `buildCompressedHistory` |
| `compression.ts` | `buildCompressedHistory` accepts optional `{ excludeLastUser }` parameter (defaults to `true` for backward compat) |

No change needed in `index.ts` — `source: next.source` is already
propagated through `runQueuedItem` → `handleRun` → `handleBridgeRun`.

The fix is opt-in: only goal kickoff runs (`data.source === 'goal_kickoff'`)
set `excludeLastUser: false`. All other runs continue with
`excludeLastUser: true`, preserving existing behavior.

## Reproduction

1. Start a new chat session in Hermes Studio
2. Send: `Explain the difference between forward proxy and reverse proxy`
3. Wait for the agent to reply with a detailed comparison
4. Send: `/goal compare the differences`
5. **Expected**: The agent analyzes the differences based on the context
   established in step 3
6. **Actual**: The agent has no recollection of the proxy discussion and
   produces an unrelated answer, trying to diff files that don't exist

## Verification

The fix was verified against a local deployment of Hermes Web UI:

1. Deployed the fix on a running instance at `v0.6.25`
2. Followed the reproduction steps:
   - "Explain the difference between forward proxy and reverse proxy" → agent replies normally
   - `/goal compare the differences` → agent now correctly references the proxy discussion and produces a meaningful comparison
3. `bridge.log` confirmed `historyMessages` now contains the full conversation context

## Validation

- TypeScript compilation: `npx tsc --noEmit` passes
- Tests: all 4 related test modules pass (37/37)
  - `session-command-plan.test.ts` (21 tests)
  - `run-chat-abort-goal.test.ts` (3 tests)
  - `run-chat-bridge-resume.test.ts` (2 tests)
  - `run-chat-queued-item.test.ts` (11 tests)


LiangYang666:
已 rebase/merge 上游最新 main（v0.6.30），仅 types.ts 存在手动冲突。本 PR 的实际改动仍集中在 goal kickoff 上下文修复，新增 goal_kickoff 到 ChatRunSource 联合类型，并合并上游 ChatCodingAgentId 类型定义。

修复后使用很久了，goal问题还是很有必要修复的

@EKKOLearnAI 帮忙看下吧

PR #1971, opened 2026-07-06
fix(bridge): use in-process BridgeServer instead of subprocess worker

### LINKED ISSUE #1970: [Bug]: bridge broker subprocess worker deadlocks on all requests; in-process worker works
## Bug Description

The Hermes Studio agent bridge consistently fails all requests with a 120s timeout. The broker spawns workers as subprocesses (WorkerProcess), but those worker processes never respond to any request — not even chat.

## Root Cause

The broker-worker subprocess communication path has a defect that prevents the worker from completing initialization or processing requests. When the same BridgeServer code is used in-process (within the broker process), all requests succeed:
- context_estimate: 0.0s
- chat (wait=False): 7.0s

## Fix

PR #1973 creates a BridgeServer directly inside the broker process and delegates all requests to it, bypassing the broken subprocess worker path entirely.

## Steps to Reproduce

1. Start Hermes Studio
2. Send any message or attempt to create a room
3. All operations time out after 120s with "Agent bridge request timed out"

## Environment

- Hermes Studio 0.6.23 (AppImage)
- All platforms (likely) but confirmed on Linux (Arch)

Love-JourneY:
## Summary

Fixes the agent bridge worker permanently hanging on all requests. The broker spawned workers as subprocesses via WorkerProcess, but those subprocesses consistently deadlocked and never responded to any request (context_estimate, chat, etc.).

## Root Cause

The subprocess worker channel has a defect that prevents the worker from processing any request. Direct in-process testing of the same BridgeServer code succeeds:
- `context_estimate`: 0.0s
- `chat` (wait=False): 7.0s

## Fix

Instead of spawning a subprocess worker, the broker now creates a BridgeServer directly in-process and delegates requests to it via `_local_server.handle()`. The subprocess worker path is retained as a fallback via the `_local_server_ready` flag.

### Changes

`packages/server/src/services/hermes/agent-bridge/python/bridge_broker.py`:
- Import BridgeServer from bridge_server
- Create `_local_server = BridgeServer(endpoint)` in `__init__`
- In `_forward()`: if `_local_server_ready` is true, call `_local_server.handle()` directly
- Fall through to subprocess worker if in-process is unavailable

## Testing

Before the fix: all bridge requests hang indefinitely → 120s timeout.
After the fix: context_estimate in 0.0s, chat in 7.0s.

## Related

Closes #1970
Building on previous bridge fixes #1949 (EPIPE) and #1964 (worker race condition).

### DIFF
--- a/packages/desktop/src/main/webui-server.ts
+++ b/packages/desktop/src/main/webui-server.ts
@@ -414,12 +414,22 @@ export async function startWebUiServer(port = DEFAULT_PORT): Promise<string> {
 
   serverProc.stdout?.on('data', (chunk: Buffer) => {
     bridgeStartup.observe(chunk)
-    process.stdout.write(`[webui] ${chunk}`)
+    try {
+      process.stdout.write(`[webui] ${chunk}`)
+    } catch {
+      /* EPIPE: parent stdout closed, ignore */
+    }
   })
+  serverProc.stdout?.on('error', () => { /* EPIPE: ignore */ })
   serverProc.stderr?.on('data', (chunk: Buffer) => {
     bridgeStartup.observe(chunk)
-    process.stderr.write(`[webui] ${chunk}`)
+    try {
+      process.stderr.write(`[webui] ${chunk}`)
+    } catch {
+      /* EPIPE: parent stderr closed, ignore */
+    }
   })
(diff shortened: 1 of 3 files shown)


Love-JourneY:
Closing — polluted diff (80 unrelated files from stale fork base). Replaced by #1973 (clean, 1 file, 13+1 lines).

PR #1973, opened 2026-07-06
fix(bridge): use in-process BridgeServer instead of subprocess worker

### LINKED ISSUE #1970: [Bug]: bridge broker subprocess worker deadlocks on all requests; in-process worker works

Love-JourneY:
## Summary

The bridge broker spawns workers as subprocesses via WorkerProcess, but those subprocesses consistently deadlock — they never respond to any request (ping, context_estimate, chat). The same BridgeServer code completes successfully in-process within 0.0s (context_estimate) to 7.0s (chat).

## Root Cause

The subprocess worker communication channel has a defect: workers initialize, print the "ready" event to stdout, and enter the accept loop, but then the broker's forwarded requests (sent over TCP to the worker) never produce a response. The worker process appears "running" from the broker's perspective but handle() never completes.

When BridgeServer is used in-process (same process, direct method call), the same handle() path completes successfully. This points to a subprocess-specific issue — likely a deadlock in the inherited thread state or file descriptor interaction between broker and worker subprocess.

## Fix

Instead of spawning a subprocess worker, the broker now creates a BridgeServer directly in-process and delegates requests via `_local_server.handle()`. The subprocess path is kept as fallback via the `_local_server_ready` flag.

## Diff

`packages/server/src/services/hermes/agent-bridge/python/bridge_broker.py`: +13, -1

- Import BridgeServer from bridge_server
- Create `_local_server = BridgeServer(endpoint)` in `__init__`
- `_forward()` delegates to `_local_server.handle()` (bypasses subprocess)
- Falls through to WorkerProcess if in-process is unavailable

## Test

```
context_estimate: ok=True, tokens=43962  (0.0s)
chat (wait=False): ok=True, run_id=...    (7.0s)
```

## Related

Closes #1970
Supersedes #1964 (same root cause, different approach — now clean diff)

### DIFF
--- a/packages/server/src/services/hermes/agent-bridge/python/bridge_broker.py
+++ b/packages/server/src/services/hermes/agent-bridge/python/bridge_broker.py
@@ -11,6 +11,7 @@ from pathlib import Path
 from typing import Any
 
 from bridge_runtime import _install_stop_signal_handlers, _jsonable
+from bridge_server import BridgeServer
 from bridge_transport import (
     WorkerProcess,
     _make_listen_socket,
@@ -28,6 +29,8 @@ class BridgeBroker:
         self.agent_root = agent_root
         self.hermes_home = hermes_home
         self._workers: dict[str, WorkerProcess] = {}
+        self._local_server = BridgeServer(endpoint)
+        self._local_server_ready = True
         self._run_profile: dict[str, str] = {}
         self._run_worker_key: dict[str, str] = {}
         self._running_run_profile: dict[str, str] = {}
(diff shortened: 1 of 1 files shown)


Love-JourneY:
Closing — this bypasses the subprocess design instead of fixing it. Will debug root cause properly.

PR #2023, opened 2026-07-10
fix(workflow): enforce exact node capability policies

### LINKED ISSUE #2022: fix(workflow): enforce node capability policy in Hermes runtime
## Problem

Workflow node roles are currently prompt-only. `WorkflowManager` sends every node through `chatRun.runAndWait()` with `source="workflow"`, but does not pass node identity or an execution policy. For Hermes-backed nodes, the bridge constructs `AIAgent` with the profile-wide enabled toolsets.

A planning node can therefore invoke file mutation, terminal, Git commit/push, or other tools even when its node prompt explicitly forbids them. Prompt wording is not a capability boundary.

## Expected behavior

A workflow node can declare an execution capability policy that is enforced by the runtime, not merely added to the prompt.

Initial minimum scope:

- persist a node-local policy in workflow snapshots;
- pass workflow/run/node identity plus the policy from `WorkflowManager` through chat-run and the bridge;
- constrain Hermes bridge sessions before context estimation and before chat execution;
- make the constrained capability set auditable in runtime/session metadata;
- add regression tests showing a constrained Plan node receives the selected toolsets while unrestricted nodes retain current behavior.

## Non-goals for this change

- changing existing workflow prompts automatically;
- altering scoped Codex/Claude Code launch permissions in the same PR;
- granting a generic terminal as a substitute for a narrowly scoped GitHub-comment capability.

## Acceptance criteria

1. A fresh Workflow run snapshot retains its declared node policy.
2. `context_estimate` and `chat` use the same policy so a session cannot be initialized with broader tool access.
3. Default/legacy nodes preserve current behavior when no policy is configured.
4. Tests cover policy propagation and the constrained bridge session construction path.

wtj-0527:
## Summary

- propagates Workflow node `executionPolicy` from the immutable run snapshot through WorkflowManager, chat-run, bridge client/server, and the Hermes agent runtime;
- keeps `allowedToolsets` as the coarse capability source, then applies `allowedTools` as an exact intersection after tool assembly; an explicit empty list remains zero-tool and unknown names cannot add capabilities;
- freezes exact-whitelist sessions against later MCP refresh so a newly available MCP cannot widen a node between turns;
- propagates `skipMemory` and `skipContextFiles` into `AIAgent` to prevent implicit memory synchronization and workspace-context injection for deterministic nodes;
- includes all policy fields in bridge-session identity: any change recreates an idle session, while a running session rejects policy changes;
- rejects policy-bearing scoped Codex/Claude nodes because that runtime cannot enforce these Hermes-specific controls.

## Security invariants

- Prompt wording is not treated as an authorization boundary.
- `allowedTools` can only reduce the tool surface resolved from `allowedToolsets`.
- `allowedTools: []` means no tools.
- Exact-whitelist sessions cannot regain tools through per-turn MCP discovery.
- Memory and context-file isolation are explicit, auditable node policy fields.

## Validation

- 27/27 targeted tests passed across WorkflowManager, bridge policy forwarding, exact tool pruning, memory/context isolation, and session recreation.
- `npm run harness:check`
- `tsc --noEmit -p packages/server/tsconfig.json`
- `python3 -m py_compile packages/server/src/services/hermes/agent-bridge/python/bridge_pool.py`
- `git diff --check`
- full `npm run build` passed before push

Standalone Codex and Claude CLI review attempts were unavailable because those local CLI sessions were not authenticated; no review result is claimed from them. GitHub Build and Playwright checks are required on the pushed head.

Closes #2022

### DIFF
--- a/docs/chat-chain-changes/2026-07-10-workflow-node-toolset-policy.md
+++ b/docs/chat-chain-changes/2026-07-10-workflow-node-toolset-policy.md
@@ -7,10 +7,14 @@ impact: Hermes-backed workflow nodes can restrict runtime toolsets per node whil
 
 # Workflow node toolset policy
 
-Workflow snapshots propagate `executionPolicy.allowedToolsets` through chat-run
-and both Agent Bridge request paths before an `AIAgent` is created. An explicit
-empty allowlist remains a zero-tool policy, and sessions are not reused across
-different policies.
+Workflow snapshots propagate `executionPolicy.allowedToolsets` and the optional
+exact `executionPolicy.allowedTools` whitelist through chat-run and both Agent
+Bridge request paths before an `AIAgent` is created. The bridge first resolves
+the allowed toolsets, then intersects the assembled Agent tool surface with the
+exact whitelist and freezes MCP refresh for that session so later connections
+cannot widen it. Policies can also set `skipMemory` and `skipContextFiles` to
+prevent persistent-memory activity and cwd/SOUL context injection. An explicit empty allowlist remains a zero-tool policy, and
+sessions are not reused across different policies.
 
 The initial enforcement applies only to Hermes-backed nodes. Workflow nodes
 using scoped Codex or Claude runtimes fail closed when configured with this
--- a/docs/chat-chain-changes/2026-07-10-workflow-node-toolset-policy.md
+++ b/docs/chat-chain-changes/2026-07-10-workflow-node-toolset-policy.md
@@ -1,0 +1,17 @@
+---
+date: 2026-07-10
+pr: 2023
+feature: Workflow node toolset policy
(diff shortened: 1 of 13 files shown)


wtj-0527:
Closing this implementation because its Workflow-specific `executionPolicy` model is no longer the direction being carried forward.

PR #2073 intentionally restores Profile / `config.yaml` defaults for Workflow nodes and strips the prerelease-only fields `allowedTools`, `allowedToolsets`, `skipMemory`, `skipContextFiles`, and Workflow-specific `executionPolicy` at read/import/run boundaries. Carrying this branch forward would reintroduce the policy surface that #2073 removes.

This closure does **not** claim that issue #2022 is resolved. If exact per-node capability isolation is still required, it should be redesigned as a separate authorization model with an explicit trust boundary rather than rebasing this patch onto the current Workflow runtime.

PR #2075, opened 2026-07-14
fix(desktop): preserve parent PATH in terminal snapshots

### LINKED ISSUE #2074: fix(desktop): preserve parent PATH in terminal snapshots
## Summary

Hermes Studio desktop currently bundles `hermes-agent==0.18.0`. Its `LocalEnvironment` starts the login shell with a parent environment whose `PATH` includes Studio/service-manager additions, but `bash -l` can replace that value before `BaseEnvironment.init_session()` writes `export -p` to the terminal snapshot.

## Reproduction

Using the packaged 0.18.0 source with an isolated HOME and a marker executable in a temporary parent-only PATH directory:

1. Construct `LocalEnvironment` with the marker directory in its supplied `PATH`.
2. Leave the directory absent from `.profile`, `.bash_profile`, and `.bashrc`.
3. Read/source the snapshot generated by the real `init_session()` path.

Observed snapshot PATH:

```text
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

The marker executable is no longer discoverable. When the same directory is exported from `.profile`, 0.18.0 retains it, confirming the remaining defect is specifically parent-only PATH entries rather than shell-init-file loading.

## Expected

The initial terminal snapshot should preserve the PATH supplied by the parent runtime while still allowing login/profile initialization to add or reorder entries. Provider credentials and other environment values must not be copied through a new broad mechanism.

## Scope

- Desktop bundled Hermes patch only.
- Preserve parent PATH for the login-shell snapshot.
- Do not change generic Studio file APIs, providers, or runtime packaging contracts.
- Add idempotency and source-anchor tests against representative Hermes 0.18.0 files.

This is a clean latest-main follow-up to #1669 rather than a reopening of the stale implementation.

wtj-0527:
## Summary

- preserve parent-runtime PATH entries that `bash -l` would otherwise discard before the desktop Hermes terminal snapshot is written;
- keep login/profile PATH ordering, appending only parent entries that are missing;
- limit the merge to POSIX, avoiding colon-based parsing of Windows `Path` values;
- remove the temporary `HERMES_INIT_PARENT_PATH` variable before `export -p` persists the snapshot;
- apply the paired `local.py` / `base.py` patch atomically and fail closed if either module or source anchor is missing.

## Root cause

Hermes Studio desktop currently bundles `hermes-agent==0.18.0`. `LocalEnvironment` passes a completed PATH into `bash -l`, but `/etc/profile` can replace it before `BaseEnvironment.init_session()` executes `export -p`. Entries injected only by the parent runtime therefore disappear even though shell-init-file entries remain available.

## Verification

- strict RED → GREEN coverage for paired application, idempotency, missing-module fail-closed behavior, and no half-applied patch when an upstream anchor changes: `3/3` PASS;
- applied the Studio patch script to a clean copy of the packaged `hermes-agent 0.18.0` source;
- patched `local.py` and `base.py` pass `py_compile`;
- real `LocalEnvironment` snapshot probe confirms:
  - snapshot initialization completes;
  - profile PATH entries retain precedence;
  - the parent-only PATH entry appears exactly once;
  - profile and parent marker binaries are both discoverable;
  - `HERMES_INIT_PARENT_PATH` is absent from the persisted snapshot;
- `npm run harness:check`: PASS;
- `npm run build`: PASS;
- `npm run test:coverage` compared with pristine latest `main` in the same environment: `39 failed / 2342 passed / 2 skipped`, `new_failures=[]` (the same 39 environment-dependent baseline failures remain).

Supersedes the stale implementation in #1669.

Closes #2074

### DIFF
--- a/packages/desktop/scripts/apply-hermes-patches.mjs
+++ b/packages/desktop/scripts/apply-hermes-patches.mjs
@@ -439,7 +439,7 @@ if (!localEnvironmentSource || !baseEnvironmentSource) {
 ` +
       `        # patch:local-parent-path-preserve — carry parent PATH through bash -l
 ` +
-      `        if login:
+      `        if login and not _IS_WINDOWS:
 ` +
       `            _path_key = _path_env_key(run_env)
 ` +
--- a/packages/desktop/scripts/apply-hermes-patches.mjs
+++ b/packages/desktop/scripts/apply-hermes-patches.mjs
@@ -43,6 +43,8 @@ const dingtalkPatchCandidates = [
 const dtPath = dingtalkPatchCandidates.find(path => existsSync(path))
 const browserToolPath = join(sitePkgs, 'tools', 'browser_tool.py')
 const sitecustomizePath = join(sitePkgs, 'sitecustomize.py')
+const localEnvironmentPath = join(sitePkgs, 'tools', 'environments', 'local.py')
+const baseEnvironmentPath = join(sitePkgs, 'tools', 'environments', 'base.py')
 if (!dtPath) {
   console.error(
     `DingTalk adapter not found. Checked:\n${dingtalkPatchCandidates.map(path => `  - ${path}`).join('\n')}`,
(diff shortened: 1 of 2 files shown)


EKKOLearnAI:
上游的问题不建议在这里打补丁

PR #2082, opened 2026-07-15
fix: attach workspace diffs to assistant turns

### LINKED ISSUE #2079: [Bug] Workspace diff cards lack coding-agent turn context
## Problem

Workspace diff checkpoints are intentionally preserved for every coding-agent turn, but the chat UI restores each checkpoint as a standalone synthetic tool message. When consecutive turns touch the same files, users see several visually identical cards with no visible turn or message context and can mistake them for duplicate rendering.

The current client associates restored changes with messages by timestamp. The persisted workspace change does not identify the user/assistant turn that produced it.

## Reproduction

1. Start a coding-agent chat session.
2. Ask the agent to modify the same two files.
3. Send two or more follow-up requests that modify those files again.
4. Reopen or refresh the session.
5. Observe multiple standalone `2 files changed` cards containing the same file names, without a visible indication that they belong to different turns.

## Expected behavior

- Preserve every per-turn workspace diff checkpoint for auditability.
- Persist an explicit link from each change to the triggering chat turn instead of inferring it from timestamps.
- Render each change as a collapsed `Changes this turn` summary attached to the corresponding assistant response.
- Keep the existing file-diff drawer and per-checkpoint patch access.
- Continue to render legacy records that do not have an explicit turn link through a safe fallback.

### Phase 1 (this issue)

- Add explicit trigger-message/turn association to coding-agent workspace changes.
- Attach the collapsed summary to the corresponding assistant response.
- Keep legacy timestamp fallback for existing rows.
- Add server/store/client regression tests.

### Out of scope

- Deleting or merging historical checkpoints.
- Summing per-turn addition/deletion counts.
- Session-level baseline-to-current net diff. That can be implemented separately because it requires a durable session baseline.

## Acceptance criteria

- Three turns modifying the same files produce three auditable checkpoints, each visibly associated with its own assistant response rather than three context-free standalone cards.
- Reloading the session preserves the same associations.
- Legacy rows without the new association remain visible.
- Diff file details continue to open by the original `change_id` and file ID.
- Relevant tests, type checks, build, and harness checks pass.

wtj-0527:
## Summary

- persist the final visible Assistant message ID on each Coding Agent workspace-diff checkpoint
- render each attributed checkpoint as a compact, collapsed **Changes this turn** summary inside that Assistant response
- keep every original checkpoint, file summary, and patch available for audit
- preserve standalone cards for legacy records and explicit associations whose Assistant message has not loaded yet
- keep ordinary tool-call and Group Chat workspace-diff behavior unchanged

Closes #2079.

## Why

Workspace diffs are intentionally incremental per Coding Agent turn. Previously, restored run checkpoints were inserted as standalone synthetic tool cards, so several turns editing the same files looked like duplicate cards even though they represented different deltas.

This change fixes the presentation and attribution instead of deleting or grouping audit history.

## Implementation

- add an additive `assistant_message_id TEXT NOT NULL DEFAULT ''` column to `workspace_run_changes`
- return the last persisted visible Assistant row ID from response flushing and pass it through checkpoint completion
- attach exact-ID checkpoints to `Message.workspaceChanges[]`
- align temporary live Assistant IDs with persisted DB IDs when the diff event arrives before terminal completion
- recompute associations after history pagination, allowing unresolved records to move under the correct Assistant once loaded
- render multiple checkpoints independently under one Assistant without overloading existing `toolChange` metadata
- add localized `chat.changesThisTurn` text to all ten locale files

## Compatibility and scope

- existing rows migrate additively and default to an empty attribution ID
- legacy and unresolved records remain visible as standalone cards with full patch access
- Group Chat's durable workspace-diff audit-message semantics are unchanged
- no checkpoints are deleted, merged by run ID, grouped by file set, or filtered to latest-only
- session baseline-to-current net diff is intentionally out of scope for this PR

## Verification

Latest base: `upstream/main` at `0dabbda394b428077ca6a19c987efb3b6f218909`.

Passed after updating to the latest base:

- `npx vitest run tests/server/workspace-diff-tracker.test.ts tests/server/response-stream-reasoning-storage.test.ts tests/server/schema-sync.test.ts tests/server/group-chat-workspace-diff.test.ts tests/client/chat-store-workspace-diff-turn.test.ts tests/client/message-item-highlight.test.ts` — **6 files, 59 tests passed**
- `npx vue-tsc -b --pretty false`
- `npx tsc --noEmit -p packages/server/tsconfig.json --pretty false`
- `npm run harness:check`
- `npm run build`
- `git diff --check upstream/main`

Manual production-build UI acceptance in an isolated state directory:

- two Assistant turns editing the same two files displayed separate nested summaries with distinct totals
- summaries were collapsed by default and expanded independently
- each same-named file opened its own original checkpoint patch (`old → turn-1`, then `turn-1 → turn-2`)
- after reopening the session, exact Assistant placement and collapsed state remained stable
- a seeded legacy row with no `assistant_message_id` remained visible as a standalone card and its file patch still opened

Full-suite note:

- an earlier `npm test` attempt on this host reported **320/324 files passed**, **2548 tests passed**, **9 failed**, **2 skipped**
- all 9 failures were confined to Agent bridge/profile/plugin discovery tests whose isolation assumptions detect the real Hermes Agent installation and bundled plugins on this host
- this branch does not modify those source modules or tests; the feature-focused suites, type checks, harness, build, and UI acceptance above pass

## Behavior

The new summary stays collapsed beneath the Assistant response until opened:

`Changes this turn · 2 files changed · +N -N`

Opening it reveals the original per-file rows and patch drawer for that exact checkpoint.

### DIFF
--- a/docs/chat-chain-changes/2026-07-15-issue-2079-workspace-diff-turn-attribution.md
+++ b/docs/chat-chain-changes/2026-07-15-issue-2079-workspace-diff-turn-attribution.md
@@ -1,0 +1,7 @@
+---
+date: 2026-07-15
+issue: "#2079"
+pr: pending
+feature: Attach Workspace Diff summaries to the Assistant response that produced them
+impact: Coding Agent checkpoints now persist the final Assistant message ID and render as collapsed per-turn summaries inside that response; legacy or not-yet-loaded records keep the existing standalone fallback without deleting audit history.
+---
--- a/docs/chat-chain-changes/2026-07-15-issue-2079-workspace-diff-turn-attribution.md
+++ b/docs/chat-chain-changes/2026-07-15-issue-2079-workspace-diff-turn-attribution.md
@@ -3,5 +3,5 @@ date: 2026-07-15
 issue: "#2079"
 pr: "#2082"
 feature: Attach Workspace Diff summaries to the Assistant response that produced them
-impact: Coding Agent checkpoints now persist the final Assistant message ID and render as collapsed per-turn summaries inside that response; legacy or not-yet-loaded records keep the existing standalone fallback without deleting audit history.
+impact: Coding Agent checkpoints and native Hermes Agent runs now persist the final Assistant message ID and render as collapsed per-turn summaries inside that response, including workspace changes made through ordinary terminal and write_file tools; legacy or not-yet-loaded records keep the existing standalone fallback without deleting audit history.
 ---
--- a/docs/chat-chain-changes/2026-07-15-issue-2079-workspace-diff-turn-attribution.md
+++ b/docs/chat-chain-changes/2026-07-15-issue-2079-workspace-diff-turn-attribution.md
@@ -1,7 +1,7 @@
 ---
 date: 2026-07-15
 issue: "#2079"
-pr: pending
+pr: "#2082"
 feature: Attach Workspace Diff summaries to the Assistant response that produced them
 impact: Coding Agent checkpoints now persist the final Assistant message ID and render as collapsed per-turn summaries inside that response; legacy or not-yet-loaded records keep the existing standalone fallback without deleting audit history.
 ---
--- a/packages/client/src/api/hermes/sessions.ts
+++ b/packages/client/src/api/hermes/sessions.ts
@@ -114,6 +114,9 @@ export interface WorkspaceRunChangeFileDetail extends WorkspaceRunChangeFileSumm
 
 export interface WorkspaceRunChangeSummary {
   change_id: string
+  room_id?: string
+  message_id?: string
+  assistant_message_id?: string
   session_id: string
   run_id: string
   source: 'run'
--- a/packages/client/src/components/hermes/chat/MessageItem.vue
+++ b/packages/client/src/components/hermes/chat/MessageItem.vue
@@ -22,7 +22,7 @@ import { useGlobalSpeech } from "@/composables/useSpeech";
 import { useVoiceSettings } from "@/composables/useVoiceSettings";
 import { speedToEdgeRate, hzToEdgePitch } from "@/utils/ttsHelpers";
 import { formatChatTimestamp } from "@/utils/chat-timestamp";
-import type { WorkspaceRunChangeFileSummary } from "@/api/hermes/sessions";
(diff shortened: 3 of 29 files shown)


wtj-0527:
## Native Hermes Agent acceptance follow-up

Commit `697b1de` extends the turn-attribution fix to native Hermes Agent bridge runs.

### Root cause

The native bridge persisted the visible Assistant response but discarded the row ID returned by `addMessage()`. The workspace checkpoint therefore completed without `assistant_message_id`, so the client could not attach the change summary to the exact Assistant turn.

### Fix

- Retain the latest persisted Assistant row ID for the active bridge run.
- Reset that ID when a new native bridge run starts.
- Pass the final ID to `completeWorkspaceRunCheckpoint()`.
- Reuse the existing `workspace_run_changes.assistant_message_id` model and UI rendering path; no native-only card implementation was added.

### Verification

- TDD regression coverage for bridge persistence and native `terminal` / `write_file` event paths.
- PR-focused regression matrix: **82/82 passed**.
- Client and server type checks: passed.
- Harness check and production build: passed.
- Real native Hermes Agent execution in an isolated production-mode UI:
  - ordinary `terminal` write produced its own collapsed **Changes this turn** card under the correct Assistant response;
  - ordinary `write_file` write produced a separate card under its correct Assistant response;
  - both expanded to the expected `+1 / -0` file diff;
  - both associations survived a full page reload;
  - persisted `assistant_message_id` values matched the corresponding final Assistant database rows.
- Full Vitest clean-environment run had 3 host-environment-sensitive command-resolution/plugin-discovery failures; the exact same 3 failures reproduce on the unmodified PR base commit `fd87dd0`, so this change introduces no new full-suite failures.
- GitHub checks for `697b1de`: **Build passed**, **Playwright passed**.

This follow-up does not merge the PR.

wtj-0527:
image

wtj-0527:
Closing in favor of the consolidated fork-internal Draft PR https://github.com/wtj-0527/hermes-studio/pull/13. That PR carries this PR's still-needed behavior (Assistant-turn workspace diff attribution) on the current upstream baseline together with the related release-carry fixes. This does not mean the change was merged upstream; it is now maintained through the single fork-side release-carry entry.

PR #2137, opened 2026-07-19
fix(workflow): preserve bridge run source across handoff

wtj-0527:
## Summary

Preserve the immutable source of each Hermes bridge run across Workflow handoff, resume, and reattachment so terminal handling applies the completed run's semantics rather than mutable state from a queued successor.

## Root cause

Workflow nodes and ordinary Chat sessions have different post-turn semantics, but terminal bridge handling previously consulted mutable Session state. By the time a run completed, that state could already describe the next queued run. A completed Workflow run could therefore enter ordinary Chat standing-goal evaluation, or a completed CLI/Global Agent run could inherit Workflow semantics.

## Changes

- Freeze an immutable source for each fresh or resumed bridge run.
- Skip ordinary Chat standing-goal evaluation only when the completed run itself is `workflow`.
- Preserve existing CLI and Global Agent standing-goal behavior.
- Derive generated ordinary Chat continuation source from the completed run, not a queued successor.
- Preserve Workflow source through reattachment and DB hydration.
- Keep every Hermes bridge run, including Workflow, behind the existing broker readiness gate.
- Update the change fragment to `pr: 2137`.

## Maintainer feedback follow-up

The observed timeout requests used the current broker endpoint with `action: "ping"`; they were not `worker_ping` requests. The logs establish a transient broker-level timeout, but do not establish that synchronous Profile-worker `goal_evaluate` blocked broker ping. The unconditional Workflow readiness bypass has therefore been removed, and a regression now asserts that Workflow Hermes runs retain the same readiness gate and outage behavior as other Hermes bridge runs.

## Verification (head `66806e3d`)

- Fresh/resume/reattach and readiness tests: 49/49 passed
- Agent Bridge Python concurrency tests: 33/33 passed, including broker ping while another broker request is blocked
- Server TypeScript check: passed
- Harness check: passed
- Production build: passed
- `git diff --check`: passed

## Scope

This does not change Workflow topology, prompts, approvals, readiness timeout, readiness error handling, or provider/model routing.

Related tracking issue: wtj-0527/hermes-studio#16.

### DIFF
--- a/docs/chat-chain-changes/2026-07-19-workflow-bridge-handoff.md
+++ b/docs/chat-chain-changes/2026-07-19-workflow-bridge-handoff.md
@@ -1,0 +1,23 @@
+---
+date: 2026-07-19
+pr: pending
+feature: Workflow bridge handoff
+impact: Hermes Workflow nodes no longer run ordinary-chat standing-goal evaluation after completion, preventing that post-turn work from blocking the next node's Agent Bridge readiness check.
+---
+
+Workflow progression is owned by the DAG scheduler, and each Workflow node uses a
+separate Session. Successful Workflow bridge runs now skip ordinary Chat's
+standing-goal continuation judge while CLI and Global Agent sessions preserve the
+existing continuation behavior.
+
+Bridge reattachment also preserves `source: workflow` in both the socket state and
+the resumed run helper, so a resumed Workflow node follows the same terminal
+semantics as a fresh node.
+
+Regression coverage forces source changes in both directions: Workflow completion
+followed by queued CLI/Global Agent work still skips the judge, while CLI/Global
+Agent completion followed by queued Workflow work still evaluates ordinary Chat
+goals and keeps the completed run's source on any generated continuation. It also
+covers empty-queue Workflow completion, synthetic completion when a bridge stream
+ends without a terminal chunk, resumed runs, and caller-side source recovery from
+a hydrated state that omits source.
--- a/docs/chat-chain-changes/2026-07-19-workflow-bridge-handoff.md
+++ b/docs/chat-chain-changes/2026-07-19-workflow-bridge-handoff.md
@@ -1,8 +1,8 @@
 ---
 date: 2026-07-19
-pr: pending
+pr: 2137
 feature: Workflow bridge handoff
-impact: Hermes Workflow nodes no longer run ordinary-chat standing-goal evaluation after completion, preventing that post-turn work from blocking the next node's Agent Bridge readiness check.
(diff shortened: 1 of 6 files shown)


EKKOLearnAI:
Thanks for the detailed regression coverage. The run-source freezing and the decision to skip ordinary Chat standing-goal evaluation for completed Workflow runs both look directionally correct.

I am not yet convinced that the readiness-bypass portion matches the current bridge architecture:

- same-profile, different-session chat runs are started on separate `hermes-bridge-run-*` threads;
- `ensureBridgeReadyForChatRun()` calls `AgentBridgeManager.ensureReady()`, which uses `AgentBridgeClient.ping()`;
- the broker handles connections concurrently, and its `ping` action returns broker-owned state without forwarding to the Profile worker;
- only `worker_ping` and other worker-routed RPCs would queue behind a synchronous `goal_evaluate` call.

Given that path, a Profile worker blocked in `goal_evaluate` should not by itself cause the current broker readiness ping to time out. Could you provide the structured log fields for the timed-out request—especially the action and endpoint—and confirm whether the reproduction used the current broker endpoint, a `worker_ping`, or a legacy/direct-worker endpoint?

Until that is established, I suggest keeping the Workflow goal-evaluation/source/resume fixes but removing the unconditional Workflow readiness bypass. Bypassing the preflight also changes real-outage behavior: the request can now wait on the normal bridge timeout, and direct `handleBridgeRun` failures do not currently pass through the readiness endpoint-redaction path.

Minor: please update the change fragment from `pr: pending` to `pr: 2137`.

wtj-0527:
Thanks — your broker/worker distinction is correct.

I rechecked the original structured logs. The two timed-out readiness requests were:

- `endpoint: ipc:///tmp/hermes-agent-bridge.sock`
- `request.action: ping`
- `timeoutMs: 1000`
- durations: `1001ms` and `1001ms`

A subsequent 75ms broker heartbeat ping also timed out, and broker ping recovered a few seconds later. There was no `worker_ping` in that failure window. These logs prove a transient broker-level timeout, but they do **not** prove that Profile-worker `goal_evaluate` blocked the broker ping. The existing broker concurrency test also confirms that broker ping remains serviceable while another broker request is blocked.

I therefore removed the unconditional Workflow readiness bypass and retained only the source/resume/reattach and Workflow standing-goal fixes. I also:

- added a regression that Workflow Hermes runs still use the normal broker readiness gate;
- updated the change fragment to `pr: 2137`;
- revised the PR title/body so they no longer make the unsupported readiness-causality claim.

Verification on head `66806e3df02a140c630daf80aaaca773cdd0f9e5`:

- fresh/resume/reattach + readiness tests: 49/49
- broker concurrency tests: 33/33
- server TypeScript: passed
- harness: passed
- production build: passed
- `git diff --check`: passed

Thank you for catching the architecture mismatch.

PR #2185, opened 2026-07-22
fix: enforce fail-closed coding agent routing

SavadiA8L:
## Summary

- add a dedicated `hermes_studio_use_coding_agent_run` MCP tool for coding-agent launches
- require exact `coding_agent_id`, `provider`, and `model` routing fields and force `source=coding_agent`
- reject empty input, generic `custom`, unsupported agents, and invalid modes before making an HTTP request
- keep the existing generic chat-run tool backward compatible
- direct coding tasks to the dedicated tool in the generated LLM prompt

## Problem

The generic chat-run MCP tool only required `input`. Some models therefore omitted the optional coding-agent routing fields, and the request silently fell back to a regular CLI/Hermes run while still looking successful to the caller.

This change makes coding-agent launches fail closed instead of silently changing execution paths.

## Contract

The dedicated tool requires `input`, `coding_agent_id` (`codex` or `claude-code`), exact `provider`, and exact `model`. It fixes `source` to `coding_agent`, defaults `mode` to `scoped`, and validates the contract both in the MCP schema and at runtime.

## Verification

After rebasing onto the current `main`:

- focused Vitest suite — 7/7 passed
- `npm run build` — passed
- `git diff --check origin/main...HEAD` — passed
- local end-to-end Grok → Codex launch verified as a real coding-agent run rather than a CLI fallback

## Compatibility

The existing `hermes_studio_use_chat_run` behavior remains available for regular Studio chat runs. The new tool only adds a stricter path for explicit coding-agent execution.

### DIFF
--- a/bin/hermes-studio-mcp.mjs
+++ b/bin/hermes-studio-mcp.mjs
@@ -576,6 +576,61 @@ function pickDefined(source, keys) {
   return picked
 }
 
+const CHAT_RUN_PAYLOAD_KEYS = [
+  'input',
+  'session_id',
+  'provider',
+  'model',
+  'model_groups',
+  'source',
+  'session_source',
+  'instructions',
+  'workspace',
+  'reasoning_effort',
+  'coding_agent_id',
+  'agent_id',
+  'mode',
+  'baseUrl',
+  'apiKey',
+  'apiMode',
+  'timeout_ms',
+  'include_events',
+]
+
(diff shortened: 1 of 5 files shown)


SavadiA8L:
Withdrawing this upstream proposal. We will keep the guard local while narrowing its scope and validating the runtime behavior.

PR #2217, opened 2026-07-25
fix custom tts voice selection

### LINKED ISSUE #2213: [Bug]: Custom TTS (OpenAI-compatible) has no Voice field — server hardcodes voice: "alloy", breaking Groq playai-tts
### Hermes Web UI Version

v0.6.33

### Hermes Agent Version

N/A (UI/server-side TTS configuration issue)

### Bug Description

When adding a TTS API with **Select Provider = "Custom TTS"** and **API compatibility = "OpenAI-compatible"**, the dialog only exposes a **Default Model** field. There is **no Voice field**.

By contrast, selecting **Volcengine Doubao TTS** does show a Voice selector.

This makes it impossible to connect any OpenAI-compatible TTS endpoint that requires a `voice` parameter. Concretely, I am trying to connect **Groq** (`https://api.groq.com/openai/v1`, model `playai-tts`), whose `/audio/speech` endpoint requires a valid `voice` (e.g. `Fritz-PlayAI`, `Arista-PlayAI`). Groq rejects the request because the UI gives me no way to set one.

Looking at the code, the plumbing for `voice` already exists end to end — only the UI input is missing:

- `packages/client/src/constants/voiceApiPresets.ts` — the `tts-custom` preset declares `capabilities: { models: true }` only, with **no `voices: true`**.
- `packages/client/src/components/hermes/settings/voice/VoiceApiFormModal.vue` — the Voice `<section>` is gated on `isDoubaoTtsPreset` (`selectedPreset?.provider === 'doubao'`), so it never renders for `tts-custom`, even though `formData.voice` exists and the save payload already forwards `voice` when non-empty.
- `packages/client/src/components/hermes/settings/voice/VoiceApiConfigurator.vue` — the edit dialog gates the Voice form item on `capabilities.voices`, so an existing Custom TTS connection also can't have a voice set after creation. (It does have a generic free-text fallback input, which would work if the capability flag were set.)
- `packages/server/src/services/hermes/tts-providers/openai.ts` — the request body hardcodes a fallback: `voice: opts.voice || 'alloy'`. `customTtsProvider` is the same `createOpenaiCompatibleTtsProvider` factory, so Custom TTS always posts `voice: "alloy"` to Groq, which is not a valid PlayAI voice.
- `packages/server/src/db/hermes/tts-settings-store.ts` — `'voice'` is already in `SETTINGS_KEYS` and is sanitized/persisted for the `custom` provider, and `mergeStoredTtsOptions` in `packages/server/src/controllers/hermes/tts.ts` already spreads it into the provider options. So persistence and request wiring work; only the input is missing.

### Steps to Reproduce

1. Go to Settings → Voice → Add TTS API.
2. Set **Select Provider** = `Custom TTS`.
3. Set **API compatibility** = `OpenAI-compatible`.
4. Enter Base URL `https://api.groq.com/openai/v1` and a Groq API key, then click **Connect & fetch models**.
5. Select model `playai-tts` in **Default Model**.
6. Observe there is no Voice field anywhere in the dialog. Save and try to synthesize.

### Expected Behavior

A **Voice** field should be available for Custom TTS / OpenAI-compatible providers (free-text or a combobox), so the configured voice is sent in the `/audio/speech` request body. Ideally the request should not silently fall back to `alloy` for a non-OpenAI endpoint where that voice does not exist.

It would also be helpful to ship a **Groq TTS preset** alongside the existing `stt-groq` preset — base URL `https://api.groq.com/openai/v1`, default model `playai-tts`. The model-probe ranking in `packages/server/src/controllers/hermes/tts.ts` already special-cases `playai`/`orpheus` names, so "Connect & fetch models" surfaces the model correctly; the voice is the only gap.

### Actual Behavior

No Voice field is rendered for Custom TTS. The server sends the hardcoded `voice: "alloy"`, which Groq's `playai-tts` rejects, so the connection cannot be used.

### Suggested fix (minimal surface)

1. `voiceApiPresets.ts` — add `voices: true` to the `tts-custom` preset capabilities (and optionally add a `tts-groq` preset mirroring `stt-groq`).
2. `VoiceApiFormModal.vue` — widen the Voice section condition from `isDoubaoTtsPreset` to also cover `kind === 'tts' && selectedPreset?.capabilities?.voices`, rendering a free-text input for non-Doubao providers. Note the Doubao-specific `@update:value="handleDoubaoVoiceUpdate"` handler must **not** run for custom providers, since it overwrites `model` with a Doubao resource ID.
3. `useVoiceApiConnections.ts` — the `custom` branch when syncing to legacy state only calls `setCustomUrl(...)` and drops model/voice; worth mirroring the `openai` branch.

### Logs / Error Messages

```shell

```

### Environment

Docker

### Node Version

_No response_

### Additional Context

For comparison, the two dialogs side by side:

| Custom TTS (OpenAI-compatible) | Volcengine Doubao TTS |
| --- | --- |
| Select Provider, API compatibility, Connection (Base URL, API Key, Connect & fetch models), Model → **Default Model** | Select Provider, Connection (Base URL, API Key), Model → Default Model (`seed-tts-2.0`), **Voice → Voice (`小何 2.0`)** |

The Custom TTS dialog ends at Default Model; the Doubao dialog has an additional Voice section below it. I'll add screenshots in a comment.

cyphercodes:
## Summary
- Expose the voice field for Custom TTS connections so OpenAI-compatible providers can use non-`alloy` voices.
- Persist the custom voice in voice settings, apply it when the custom provider is selected, and pass it through group-chat playback.
- Add regression coverage for applying a saved custom TTS voice.

Fixes #2213

## Validation
- `npm run test -- tests/client/use-voice-api-connections-mimo.test.ts tests/server/openai-tts-provider.test.ts`
- `npm run harness:check`
- `NODE_OPTIONS=--max-old-space-size=4096 npx vue-tsc -b --pretty false`
- `NODE_OPTIONS=--max-old-space-size=4096 npm run build`

### DIFF
--- a/docs/chat-chain-changes/2026-07-25-custom-tts-voice.md
+++ b/docs/chat-chain-changes/2026-07-25-custom-tts-voice.md
@@ -1,0 +1,13 @@
+---
+date: 2026-07-25
+pr: pending
+feature: Custom OpenAI-compatible TTS voice selection
+impact: Custom TTS playback can persist and send a configured voice instead of always relying on the provider default.
+---
+
+Custom OpenAI-compatible TTS connections now expose the same voice field used by
+other speech providers. Selecting the custom provider copies the stored voice
+into the legacy playback state, and group-chat playback includes that voice in
+the server synthesize request.
+
+Chat text, session persistence, and message ordering are unchanged.
--- a/packages/client/src/components/hermes/group-chat/GroupMessageItem.vue
+++ b/packages/client/src/components/hermes/group-chat/GroupMessageItem.vue
@@ -457,6 +457,7 @@ function playSpeech(content: string, autoplay = false) {
             provider: 'custom' as const,
             baseUrl: voiceSettings.customUrl.value,
             apiKey: voiceSettings.customApiKey.value || undefined,
+            voice: voiceSettings.customVoice.value,
         }
         if (autoplay) void speech.openaiPlay(props.message.id, content, options).catch(handleAutoplayTtsError)
         else speech.openaiToggle(props.message.id, content, options)
--- a/packages/client/src/composables/useVoiceApiConnections.ts
+++ b/packages/client/src/composables/useVoiceApiConnections.ts
@@ -87,6 +87,7 @@ export function useVoiceApiConnections() {
 
     if (connection.provider === 'custom') {
       vs.setCustomUrl(connection.baseUrl || stringSetting(settings, 'baseUrl'))
+      vs.setCustomVoice(connection.voice || stringSetting(settings, 'voice') || vs.customVoice.value)
       return
     }
 
--- a/packages/client/src/composables/useVoiceSettings.ts
+++ b/packages/client/src/composables/useVoiceSettings.ts
@@ -19,6 +19,7 @@ export interface VoiceSettingsData {
   // Custom endpoint (OpenAI-compatible)
   customUrl: string
   customApiKey: string
+  customVoice: string
 
   // Edge TTS
   edgeUrl: string
@@ -83,6 +84,7 @@ const DEFAULT: VoiceSettingsData = {
 
   customUrl: '',
   customApiKey: '',
+  customVoice: 'alloy',
 
   edgeUrl: '',
   edgeVoice: 'zh-CN-XiaoxiaoNeural',
@@ -147,6 +149,7 @@ const openaiVoice = ref<string>(load().openaiVoice)
 // Custom
 const customUrl = ref<string>(load().customUrl)
 const customApiKey = ref<string>(load().customApiKey)
+const customVoice = ref<string>(load().customVoice)
 
 // Edge TTS
 const edgeUrl = ref<string>(load().edgeUrl)
@@ -176,7 +179,7 @@ const doubaoStylePrompt = ref<string>(load().doubaoStylePrompt)
 // Auto-persist on change
 watch(
   [provider, webspeechVoice, openaiApiKey, openaiBaseUrl, openaiModel, openaiVoice,
-   customUrl, customApiKey, edgeUrl, edgeVoice, edgeRate, edgePitchHz,
+   customUrl, customApiKey, customVoice, edgeUrl, edgeVoice, edgeRate, edgePitchHz,
    mimoApiKey, mimoAuthMode, mimoBaseUrl, mimoModel, mimoVoice, mimoVoiceDesignDesc,
    mimoVoiceCloneDataUri, mimoVoiceCloneFileName, mimoVoiceCloneFormat, mimoStylePrompt,
    doubaoApiKey, doubaoBaseUrl, doubaoModel, doubaoVoice, doubaoStylePrompt],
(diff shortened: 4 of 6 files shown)


EKKOLearnAI:
Thanks for working on this. There is still a blocking gap in the primary #2213 flow: `VoiceApiFormModal.vue` continues to render the Voice section only when `isDoubaoTtsPreset` is true. Adding `capabilities.voices` to `tts-custom` enables the edit configurator, but the **Add TTS API → Custom TTS** dialog still has no Voice input, so a new Groq/PlayAI connection cannot be configured with `Fritz-PlayAI` (or another required voice) during creation.

Could you please:

1. Render a free-text Voice input in `VoiceApiFormModal.vue` for TTS presets with `capabilities.voices`, while keeping the Doubao-specific selector and `handleDoubaoVoiceUpdate` behavior limited to Doubao.
2. Add a component regression test that selects `tts-custom`, verifies the Voice field is visible, enters a value, and verifies it is included in the save payload.

There is also an upgrade-path risk with the new local `customVoice` default of `alloy`: group-chat playback now sends that value explicitly, which can override a correctly stored custom voice for an existing user whose older local settings do not yet contain `customVoice`. Please either avoid sending a default custom voice until the stored provider setting has been loaded, or initialize it from the saved server setting before playback.

twuijri:
Reporter here (filed #2213). @cyphercodes thanks for picking this up — I've implemented the three review items on top of your commit and opened them **against your branch**, not against `main`, so this can land as part of this PR under your name: cyphercodes/hermes-studio#1

Merge it if it looks right, close it if you'd rather implement it yourself. I didn't want to open a competing PR.

**1. Voice field in the Add dialog.** `VoiceApiFormModal` now renders the Voice section for any TTS preset that declares `capabilities.voices`, instead of Doubao only. Doubao keeps its curated `NSelect` and the resource-id mapping via `handleDoubaoVoiceUpdate`; every other preset (including Custom TTS) gets a free-text `NInput`, so a Groq PlayAI/Orpheus voice can be set at creation time.

One useful finding: `VoiceApiConfigurator` already had a free-text voice input gated on the same capability flag, so your preset change fixed the **edit** path — the Add dialog was the only remaining gap.

**2. Upgrade-path risk.** `customVoice` now defaults to `''` instead of `'alloy'`, and `GroupMessageItem` omits `voice` when empty. Otherwise a client with older local settings picks up `'alloy'` from `{ ...DEFAULT, ...parsed }` and sends it unconditionally, overriding a correctly stored provider voice. Concrete case: my setup stores `voice: "fahad"` for Groq `canopylabs/orpheus-arabic-saudi`, which the provider rejects when the client forces `alloy`.

**3. Component regression tests.** `tests/client/voice-api-form-modal-voice.test.ts` — five cases: field hidden before a preset is chosen, free-text input for Custom TTS, Doubao keeps its picker, voice included in the save payload, voice omitted when empty.

**Verification:** 5/5 new tests pass, `vue-tsc --noEmit` clean, `npm run harness:check` passes. Full suite shows 12 failures in `studio-mcp-autoinject` and `use-speech-tts` — identical on this branch without my commit **and** on `main` before it, so they're pre-existing and unrelated. No new i18n keys (reuses `settings.voice.voice`) per the locale rule in `AGENTS.md`, and I updated the existing `docs/chat-chain-changes` fragment rather than adding a second one.

Two things I deliberately left out to keep the scope tight, happy to follow up if @EKKOLearnAI prefers: the server-side hardcoded `voice: "alloy"` fallback in `openai.ts`, and single-chat playback, which doesn't appear to pass the voice the way group chat now does.

PR #2258, opened 2026-07-28
fix(voice): mirror Web UI STT provider into the Hermes Agent profile config

### LINKED ISSUE #2257: Web UI voice settings are not mirrored to the Hermes Agent profile, so gateway platforms never use the configured STT provider
### Summary

Voice settings configured in **Settings → Voice** only affect browser transcription. Hermes Agent gateway platforms (Telegram, Discord, …) read a **different** configuration store, so a provider configured in Studio never reaches them.

Two separate stores exist today:

| Path | Who transcribes | Configuration source |
| --- | --- | --- |
| Web UI microphone | Studio server, `POST /api/hermes/stt/transcribe` | Studio settings DB (Settings → Voice) |
| Telegram / Discord voice message | Hermes Agent gateway | profile `config.yaml` `stt:` block + profile `.env` |

Result: I configure Groq STT in Studio, the built-in **Test** succeeds, and Telegram voice messages still don't use Groq — Hermes resolves `stt.provider` to `local` because the block was never written. Nothing in the UI indicates that a second configuration is required.

### Reproduce

1. Settings → Voice → add an STT provider with base URL `https://api.groq.com/openai/v1` and a Groq key.
2. Press **Test** — transcription succeeds in the browser.
3. Send a voice message to the profile's Telegram bot.
4. Hermes transcribes with `local` (or reports a generic failure). `~/.hermes/profiles/<profile>/config.yaml` has no `stt:` block and `.env` has no `GROQ_API_KEY`.

### Why this belongs in Studio

Studio already owns this kind of mirroring — this is a missing case, not a new architecture:

- `controllers/hermes/providers.ts:302-369` — saving a model provider writes `custom_providers` into `config.yaml` **and** calls `saveEnvValueForProfile` for the API key, in the same handler.
- `controllers/hermes/config.ts:438-456` — proxy settings are written to `.env`, then the gateway is restarted.
- `controllers/hermes/config.ts:636-691` — platform credentials (including `TELEGRAM_BOT_TOKEN`) are written to `.env` and deliberately kept out of `config.yaml`.

The helpers are already in place: `updateConfigYamlForProfile` (`services/config-helpers.ts:104`), `saveEnvValueForProfile` (`services/config-helpers.ts:153`), `restartGatewayForProfile` (`services/hermes/gateway-autostart.ts:645`).

Voice is the only settings group that stops at the Studio DB.

### Proposed fix

When an STT provider is saved and its base URL maps unambiguously to a Hermes provider, mirror two values into the profile:

- `.env` ← the credential, under the env var Hermes reads (`GROQ_API_KEY`)
- `config.yaml` ← `stt.enabled: true`, `stt.provider: groq`

Key names follow the Hermes Agent reference (`website/docs/user-guide/features/voice-mode.md`):

```yaml
stt:
  enabled: true
  provider: "local"   # "local" | "groq" | "openai" | "mistral" | "xai"
```

Scoped to Groq for the first pass so a generic OpenAI-compatible endpoint is never silently relabelled; other providers can be added to one table once the approach is agreed. Existing user-written keys under `stt:` (e.g. `local.model`, `groq.language`) are preserved, and a write failure is logged without failing the settings request.

Implementation with tests: (PR linked below.)

### One question for the maintainer

`AGENTS.md` says *"Keep Hermes Agent state separate from Web UI state."* My reading is that this forbids storing **Web UI state inside Hermes**, not writing Hermes' own configuration — which Studio already does for model providers, proxy, and platform credentials. If you read it more strictly, I'm happy to move the mirror behind an explicit opt-in (a toggle or an "Apply to Hermes Agent" action) instead of running on save.

### Environment

- Studio v0.6.33, Hermes Agent 0.19, Docker deployment

twuijri:
Fixes #2257.

Voice was the only settings group that stopped at the Studio DB: it served `POST /api/hermes/stt/transcribe` (browser only), while Hermes Agent gateway platforms read a separate `stt:` block from the profile `config.yaml` plus credentials from the profile `.env`. A provider configured in **Settings → Voice** therefore never reached Telegram, Discord or any other gateway platform, and Hermes kept resolving `stt.provider` to `local`.

### What this does

New service `packages/server/src/services/hermes/voice-config-sync.ts`, called from `saveSettings` in `controllers/hermes/stt.ts`:

- `.env` ← the credential under the env var Hermes reads (`GROQ_API_KEY`), via the existing `saveEnvValueForProfile`
- `config.yaml` ← `stt.enabled: true` and `stt.provider`, via the existing `updateConfigYamlForProfile`

Key names follow the Hermes Agent reference (`website/docs/user-guide/features/voice-mode.md`).

This is the same pattern Studio already uses elsewhere — `providers.ts:302-369` (model providers → `custom_providers` + API key in `.env`), `config.ts:438-456` (proxy → `.env`), `config.ts:636-691` (platform credentials → `.env`).

### Deliberate constraints

- **Groq only for the first pass.** `resolveHermesSttTarget` matches on host, so a generic OpenAI-compatible endpoint is never silently relabelled as a Hermes provider. Adding providers later is one table entry each.
- **Look-alike hosts rejected** — `api.groq.com.evil.test` does not match (covered by a test).
- **Stored-secret placeholder ignored** — `[stored]` never overwrites a real key in `.env`.
- **User keys preserved** — existing `stt.local.model`, `stt.groq.language`, etc. are kept; only `enabled` and `provider` are set.
- **Never fails the request** — a write error is logged and the settings response is unchanged.
- **Browser path untouched** — Web UI transcription keeps reading the Web UI settings store.
- **No gateway restart** — left out on purpose; happy to add `restartGatewayForProfile` if you prefer the change to apply without a manual restart.

### Verification

- `tests/server/voice-config-sync.test.ts` — 10 new tests (host mapping, look-alike rejection, malformed input, placeholder secret, `stt:` block creation/merge/replacement, unrelated config untouched): all pass
- `tsc --noEmit -p packages/server/tsconfig.json` clean
- `npm run harness:check` passes
- `tests/server` suite: 1877 passed, 11 failures in `studio-mcp-autoinject` that are pre-existing on `main` and unrelated to this change

### Open question

`AGENTS.md` says *"Keep Hermes Agent state separate from Web UI state."* I read that as forbidding Web UI state stored **inside** Hermes, not writing Hermes' own configuration — which Studio already does in the three places above. If you read it more strictly, I'll move the mirror behind an explicit opt-in (a toggle, or an "Apply to Hermes Agent" action) instead of running on save; say the word and I'll push that instead.

### DIFF
--- a/docs/chat-chain-changes/2026-07-26-stt-config-sync.md
+++ b/docs/chat-chain-changes/2026-07-26-stt-config-sync.md
@@ -1,0 +1,18 @@
+---
+date: 2026-07-26
+pr: pending
+feature: Mirror Web UI STT provider into the Hermes Agent profile config
+impact: Saving a Groq STT provider in Settings -> Voice now also configures Hermes Agent, so gateway platforms transcribe voice messages with the same provider instead of falling back to local.
+---
+
+Saving an STT provider whose base URL maps unambiguously to a Hermes Agent
+provider now mirrors two values into the profile: the credential is written to
+the profile `.env` under the env var Hermes reads, and `stt.enabled` /
+`stt.provider` are written to the profile `config.yaml`. Only Groq is mapped for
+now, so a generic OpenAI-compatible endpoint is never silently relabelled.
+
+The mirror never fails the request: a write error is logged and the settings
+response is unchanged. Web UI transcription keeps using the Web UI settings
+store, so browser behaviour is unchanged.
+
+Chat text, session persistence, and message ordering are unchanged.
--- a/packages/server/src/controllers/hermes/stt.ts
+++ b/packages/server/src/controllers/hermes/stt.ts
@@ -20,6 +20,7 @@ import { config } from '../../config'
 import { SttProviderConfigError, transcribeWithProvider } from '../../services/hermes/stt-providers'
 import { SttNoSpeechDetectedError } from '../../services/hermes/stt-providers/types'
 import { logger } from '../../services/logger'
+import { syncSttProviderToHermesConfig } from '../../services/hermes/voice-config-sync'
 import { getActiveGlobalAgentServer } from '../../services/global-agent/server'
 import { MCU_TTS_SAMPLE_RATE, mcuPromptText, mcuPromptUrl } from '../../services/hermes/mcu-prompts'
 
@@ -345,6 +346,15 @@ export async function saveSettings(ctx: Context) {
     const activeProvider = body?.activeProvider === undefined
       ? saveActiveSttProvider(profile, storedProvider)
       : saveActiveSttProvider(profile, assertActiveSttProvider(String(body.activeProvider)))
+
+    // Mirror the provider into the Hermes Agent profile config so gateway
+    // platforms (Telegram, Discord, ...) transcribe with the same provider.
+    const settingsBody = body?.settings as Record<string, unknown> | undefined
+    const secretsBody = body?.secrets as Record<string, unknown> | undefined
+    await syncSttProviderToHermesConfig(profile, {
+      baseUrl: settingsBody?.baseUrl,
+      apiKey: secretsBody?.apiKey,
+    })
 
     ctx.body = { setting, activeProvider }
   } catch (error) {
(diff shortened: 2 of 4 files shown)


EKKOLearnAI:
https://github.com/EKKOLearnAI/hermes-studio/pull/2260

PR #2304, opened 2026-07-31
fix(cli): find the Hermes CLI when PATH does not

### LINKED ISSUE #2129: [Bug]: systemd unit for hermes-web-ui missing PATH — 'hermes logs' fails after reboot
## Description

After a VM reboot, the **Monitoring** tab in Hermes Web UI cannot display logs.

## Root Cause

The systemd user service unit does not set `Environment=PATH`. Systemd user services inherit a minimal PATH which does not include `~/.local/bin` where `hermes` CLI lives.

## Impact

Any user running on a headless VPS loses monitoring logs after every reboot.

## Workaround

Add to the systemd unit:
```ini
[Service]
Environment=PATH=$HOME/.local/bin:$HOME/.hermes/node/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/snap/bin
```
Then:
```bash
systemctl --user daemon-reload
systemctl --user restart hermes-web-ui.service
```

## Suggested Fix

The installer should detect `hermes` location and inject correct `Environment=PATH` into the systemd unit template.

twuijri:
### The problem

The Web UI invokes the CLI by bare name (`hermes`), resolved through PATH. A `systemd --user` service inherits a minimal PATH that usually leaves out `~/.local/bin` — exactly where the CLI installs itself.

So after a reboot, every hermes call from the server fails with ENOENT on a machine where hermes is plainly installed. The Monitoring tab shows no logs, and the gateway and profile calls fail the same way. #2129 works around it by hand-editing the unit file.

### The fix

`resolveHermesBin` keeps its order of precedence and gains a last resort:

1. `HERMES_BIN` — unchanged, still authoritative
2. PATH — unchanged, still preferred when it resolves
3. **new:** the known install locations — `~/.local/bin`, `~/.hermes/node/bin`, `~/.local/share/hermes/bin`, `/usr/local/bin`, `/opt/homebrew/bin`

Windows resolution is untouched (it returns the bare name as before, and the bundled-python path logic is unchanged). The answer is cached for the life of the process, so this costs a handful of `access()` calls once.

Two copies of the resolver had drifted into `hermes-cli.ts` and `gateway-autostart.ts`, neither of which would have picked up the fallback; both now use the shared one.

### Tests

`tests/server/hermes-bin-resolution.test.ts` covers each branch with a real executable in a temporary HOME: PATH hit, `~/.local/bin` fallback, `~/.hermes` fallback, `HERMES_BIN` override, and nothing installed.

`npm run harness:check` passes. The whole `tests/server` suite passes except `studio-mcp-autoinject.test.ts`, which fails identically on a clean `main` in this environment — unrelated to this change.

Fixes #2129

### DIFF
--- a/packages/server/src/services/hermes/gateway-autostart.ts
+++ b/packages/server/src/services/hermes/gateway-autostart.ts
@@ -8,7 +8,7 @@ import { logger } from '../logger'
 import { getHermesBaseDir, getProfileDir, listProfileNamesFromDisk } from './hermes-profile'
 import { retireManagedGatewayForProfile, startGatewayRunManaged } from './gateway-runner'
 import { parseGatewayStatusesFromProfileList } from './profile-list-parser'
-import { execHermesWithBin } from './hermes-process'
+import { execHermesWithBin, resolveHermesBin } from './hermes-process'
 
 const execFileAsync = promisify(execFile)
 const GATEWAY_RUNTIME_FILES = ['gateway.pid', 'gateway.lock', 'gateway_state.json'] as const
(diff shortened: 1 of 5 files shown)


twuijri:
Closing this — changing how the CLI binary is resolved is closer to the system than I want to reach into from outside. Leaving #2129 to you.

PR #2325, opened 2026-08-02
fix(bridge): honor model-specific reasoning overrides

### LINKED ISSUE #2306: [Bug]: Agent Bridge ignores per-model reasoning_overrides for fresh default sessions
## Summary

When Hermes Agent is configured with a global reasoning effort plus a per-model override, a fresh default-valued chat session in Hermes Studio ignores the per-model override and uses the global effort instead.

This happens in the Studio Agent Bridge default-agent construction path. It is not caused by a persisted old session or by the per-session reasoning control when that control is left at its default.

## Environment

- Hermes Web UI: `0.6.33` (active installed package)
- Hermes Agent: `v0.19.0`
- Node.js: `v24.16.0`
- OS: Linux
- Studio source audited: `v0.6.33`, `v0.6.35`, and current `main`
- Provider/model names used for reproduction: `gpt-5.6-luna` and `gpt-5.6-sol`

## Configuration

```yaml
agent:
  reasoning_effort: xhigh
  reasoning_overrides:
    gpt-5.6-sol: xhigh
    gpt-5.6-luna: max
```

## Steps to reproduce

1. Configure a global effort and a model-specific override as shown above.
2. Open Hermes Studio.
3. Leave the chat reasoning control at its default value (`Default (config.yaml)`). Do not select a per-session effort.
4. Create a brand-new chat session and send the first message using `gpt-5.6-luna`.
5. Inspect the resulting Hermes session runtime configuration, for example the persisted `sessions.model_config.reasoning_config.effort` field in `state.db`.
6. Compare it with Hermes core's model-aware resolver for the same config and model.

## Expected behavior

With no explicit per-session reasoning override, the Studio Agent Bridge should use Hermes Agent's normal model-aware configuration resolution:

```text
gpt-5.6-luna -> max
gpt-5.6-sol  -> xhigh
```

The Studio default control should remain default/inherit behavior. Users should not have to manually select `Max` in every new Luna session, and switching models without an explicit session override should resolve the corresponding model override.

## Actual behavior

A fresh Luna session is created with:

```json
{
  "reasoning_config": {
    "enabled": true,
    "effort": "xhigh"
  }
}
```

The installed Studio Bridge loader returns the global value:

```text
Bridge loader: xhigh
Hermes core model-aware resolver for gpt-5.6-luna: max
```

A real fresh session used for verification had zero history at creation time and persisted `xhigh` on its first run, so this is reproducible without resuming an older conversation.

## Source evidence

The Studio per-session selector is intentionally an explicit override:

- `ChatInput.vue` defines the default option as an empty value, documented as `Default (config.yaml)`:
  https://github.com/EKKOLearnAI/hermes-studio/blob/841ab4a8e6a56d09407855dc10d6afa2d1bb8125/packages/client/src/components/hermes/chat/ChatInput.vue#L51-L67
- The client stores a non-empty selection in browser storage keyed by session ID and does not copy it as a global default:
  https://github.com/EKKOLearnAI/hermes-studio/blob/841ab4a8e6a56d09407855dc10d6afa2d1bb8125/packages/client/src/stores/hermes/chat.ts#L4609-L4643
- The existing per-session feature explicitly documents that its default option falls back to the global `agent.reasoning_effort`; it does not define model-specific override resolution:
  https://github.com/EKKOLearnAI/hermes-studio/pull/1417

The bug is in the Python Agent Bridge default path:

- `_load_reasoning_config()` parses only `agent.reasoning_effort` and does not call Hermes's model-aware resolver:
  https://github.com/EKKOLearnAI/hermes-studio/blob/841ab4a8e6a56d09407855dc10d6afa2d1bb8125/packages/server/src/services/hermes/agent-bridge/python/bridge_runtime.py#L897-L905
- Agent construction calls it without passing the already-resolved model:
  https://github.com/EKKOLearnAI/hermes-studio/blob/841ab4a8e6a56d09407855dc10d6afa2d1bb8125/packages/server/src/services/hermes/agent-bridge/python/bridge_pool.py#L299-L317

The same relevant Bridge code is present in the audited `v0.6.35` source and current `main` at the time of filing.

## Proposed fix

Delegate to Hermes Agent's shared resolver and pass the selected model:

```python
def _load_reasoning_config(model: str = "") -> dict[str, Any] | None:
    _ensure_agent_imports()
    try:
        from hermes_constants import resolve_reasoning_config
        return resolve_reasoning_config(_load_cfg(), model)
    except Exception:
        return None
```

At Agent construction:

```python
reasoning_config=_load_reasoning_config(resolved_model)
```

This preserves the existing behavior for the global default while enabling model-specific overrides. An explicit non-empty Studio per-session value should continue to take precedence for that run/session.

## Verification of the proposed fix

A local candidate patch was tested against the current configuration:

```text
patched Bridge + gpt-5.6-luna -> max
patched Bridge + gpt-5.6-sol  -> xhigh
```

Focused Studio tests passed in an isolated checkout:

```text
2 test files passed
19 tests passed
```

Suggested regression coverage:

- Bridge default construction resolves a Luna model to its `agent.reasoning_overrides` value.
- A model without an override still receives the global `agent.reasoning_effort`.
- An explicit per-session `reasoning_effort` continues to override the Agent default.
- The default Studio control omits `reasoning_effort` from the run payload.

## Impact

Users who rely on `agent.reasoning_overrides` receive the wrong reasoning effort in every fresh Studio session unless they manually select the effort. This is especially confusing because the same configuration resolves correctly through Hermes Agent's core model-aware path, and the Studio UI labels the default as `Default (config.yaml)`.

This is a Studio-only Bridge issue; restarting the Gateway does not change the affected code path.

## Related work

- Related but distinct: [#1417](https://github.com/EKKOLearnAI/hermes-studio/pull/1417) added the per-session reasoning effort selector. This report concerns the default Bridge resolution when the selector is left at its documented default.
- Related broader metadata discussion: [#1310](https://github.com/EKKOLearnAI/hermes-studio/issues/1310).

qdivan:
## Summary

- pass the resolved model into the Agent Bridge reasoning-config loader so fresh Studio agents use Hermes Agent's per-model `reasoning_overrides`
- retain the previous global `reasoning_effort` fallback for older Hermes runtimes
- preserve explicit per-run reasoning overrides and restore the model-aware default after both successful and failed runs
- add regression coverage for model-aware defaults, legacy fallback, resolver errors, and override restoration

Fixes #2306

## Validation

- `npm run test -- tests/server/agent-bridge-reasoning-effort.test.ts tests/server/agent-bridge-profile-env.test.ts` — 22/22 passed
- `npm run build` — passed
- `git diff --check` — passed
- independent Codex audit: candidate quality PASS; patch quality PASS
- publication-refresh replay: tracked diff and docs fragment byte-identical to the audited patch on canonical `main` at `6c14854a9a00433756f7c7fd2d303c39f46d74c7`

## Duplicate Check

Rechecked live on 2026-08-02 before publication:

- issue #2306 is open, with no labels or comments indicating ownership or a competing implementation
- exact issue-number PR search returned no matches
- model-aware/reasoning PR search found no patch touching the five changed paths
- recent canonical history and the refreshed `main` do not contain the fix

## Browser / Playwright

Not applicable. This patch changes the server-side Python Agent Bridge reasoning configuration and its unit-level regression coverage; it does not change a browser-visible UI flow.

### DIFF
--- a/docs/chat-chain-changes/2026-08-02-model-aware-reasoning-defaults.md
+++ b/docs/chat-chain-changes/2026-08-02-model-aware-reasoning-defaults.md
@@ -1,0 +1,9 @@
+---
+date: 2026-08-02
+pr: pending
+feature: Model-aware Agent Bridge reasoning defaults
+impact: Fresh Studio agents now inherit Hermes per-model reasoning overrides before the global reasoning effort.
+---
+
+Explicit per-run reasoning effort remains the highest-priority override and is
+restored to the model-aware constructor default after the run.
--- a/docs/chat-chain-changes/2026-08-02-model-aware-reasoning-defaults.md
+++ b/docs/chat-chain-changes/2026-08-02-model-aware-reasoning-defaults.md
@@ -1,9 +1,9 @@
 ---
 date: 2026-08-02
-pr: pending
+pr: 2325
 feature: Model-aware Agent Bridge reasoning defaults
-impact: Fresh Studio agents now inherit Hermes per-model reasoning overrides before the global reasoning effort.
+impact: Fresh and successfully model-switched Studio agents now inherit Hermes per-model reasoning overrides before the global reasoning effort.
 ---
 
 Explicit per-run reasoning effort remains the highest-priority override and is
-restored to the model-aware constructor default after the run.
+restored to the current model-aware default after the run.
--- a/packages/server/src/services/hermes/agent-bridge/python/bridge_pool.py
+++ b/packages/server/src/services/hermes/agent-bridge/python/bridge_pool.py
@@ -313,7 +313,7 @@ class AgentPool:
                     credential_pool=runtime.get("credential_pool"),
                     quiet_mode=True,
                     verbose_logging=False,
-                    reasoning_config=_load_reasoning_config(),
+                    reasoning_config=_load_reasoning_config(resolved_model),
                     service_tier=_load_service_tier(),
                     enabled_toolsets=_load_enabled_toolsets(),
                     platform=_bridge_platform(),
--- a/packages/server/src/services/hermes/agent-bridge/python/bridge_pool.py
+++ b/packages/server/src/services/hermes/agent-bridge/python/bridge_pool.py
@@ -458,6 +458,10 @@ class AgentPool:
         if not callable(switch_model):
             raise RuntimeError("loaded agent does not support switch_model")
 
+        with _profile_env(target_profile):
+            _refresh_worker_profile_env()
+            reasoning_config = _load_reasoning_config(requested_model)
+
         switch_model(
             new_model=requested_model,
             new_provider=resolved_provider,
@@ -465,6 +469,7 @@ class AgentPool:
             base_url=runtime.get("base_url") or "",
             api_mode=runtime.get("api_mode") or "",
         )
+        session.agent.reasoning_config = reasoning_config
         if resolved_provider.lower() == "moa":
             self._install_moa_reference_callback(session)
         session.config.update({
--- a/packages/server/src/services/hermes/agent-bridge/python/bridge_runtime.py
+++ b/packages/server/src/services/hermes/agent-bridge/python/bridge_runtime.py
@@ -894,15 +894,19 @@ def _log_worker_startup_context(profile: str | None) -> None:
         })
 
 
-def _load_reasoning_config() -> dict[str, Any] | None:
+def _load_reasoning_config(model: str = "") -> dict[str, Any] | None:
     _ensure_agent_imports()
     try:
-        from hermes_constants import parse_reasoning_effort
+        from hermes_constants import resolve_reasoning_config
+    except ImportError:
+        try:
+            from hermes_constants import parse_reasoning_effort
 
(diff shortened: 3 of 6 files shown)


EKKOLearnAI:
Thanks — the fresh-session path now resolves the model-specific override correctly. I think one loaded-session path is still missing, though.

`_switch_loaded_session_model()` calls `session.agent.switch_model(...)` and updates `session.config`, but it never recomputes `session.agent.reasoning_config`. For example, a session created on Luna with `reasoning=max` and then switched to Sol can continue using the Luna value `max` instead of the configured Sol value `xhigh` whenever the Studio session reasoning control is left at Default. An explicit per-run/session override masks this, but the default path remains stale.

Since #2306 also expects model switches without an explicit session override to resolve the override for the new model, could you refresh the reasoning config from `_load_reasoning_config(requested_model)` after a successful model switch and add a Luna → Sol regression test?

qdivan:
Thanks — addressed in `0dd504e`: loaded-session model switches now recompute `reasoning_config` for the requested model, with Luna → Sol regression coverage. Follow-up `9747366` fixes the concurrency test harness stub. The exact current head has green Build and Playwright checks; re-review appreciated.

PR #2329, opened 2026-08-03
Refresh MiniMax TTS defaults and regional presets

octo-patch:
Reason: Refresh the MiniMax TTS preset to use the newer default model and expose both regional endpoints.

Updated the MiniMax TTS preset to add a China-region option, keep the newer speech-2.8-hd default, and preserve the same TTS payload behavior in the server path.

Checks:
- `NODE24=/Users/kape/.npm/_npx/387698761821791d/node_modules/.bin/node ./node_modules/vitest/vitest.mjs run tests/client/voice-api-presets-custom-voice.test.ts tests/server/hermes-cloud-voice-providers.test.ts`
- Secret scan passed


EKKOLearnAI:
Thanks for the focused update. The MiniMax model/endpoint changes and tests look good, and the current CI is green.

One repository requirement remains before merge: the new user-visible `MiniMax TTS (China)` label is hardcoded and has no `labelKey`. Every existing voice preset uses localized labels, and this repository requires user-facing strings in every locale. Please add a dedicated locale key (for example `settings.voice.presetMinimaxCnTtsLabel`) to all locale files and reference it from the China preset.

After that small i18n fix, this looks ready.

octo-patch:
Thanks for the review. I added the China preset label key across every locale and extended the preset test, then pushed a new commit. The focused suite passes all 5 tests.

PR #2332, opened 2026-08-03
Add MiniMax voice clone support

octo-patch:
Reason: MiniMax voice cloning is exposed by the provider layer but is not wired into the MiniMax preset.

This change adds MiniMax voice clone support for global and regional endpoints, including reference-audio upload, clone creation, synthesis with the returned voice ID, and the client configuration flow for uploading an MP3 or WAV reference clip. It keeps the change limited to voice cloning.

Checks:
- npm test -- --run tests/server/hermes-cloud-voice-providers.test.ts tests/client/use-speech-tts.test.ts tests/client/use-voice-api-connections-mimo.test.ts
- npx tsc --noEmit -p packages/server/tsconfig.json
- npx vue-tsc -b tsconfig.app.json tsconfig.node.json tsconfig.website.json
- git diff --check


EKKOLearnAI:
Thanks for working on MiniMax voice cloning. The current CI is green, but I found several functional blockers that should be addressed before merge:

1. `resolveMiniMaxVoiceId()` uploads and calls `/voice_clone` from every `minimaxTtsProvider.synthesize()` invocation. Voice cloning is a setup operation: after creating the clone, later TTS calls should reuse the stored `voice_id`. MiniMax requires custom voice IDs to be unique, so the second playback will attempt to create the same ID again and can fail. Please move cloning to a one-time lifecycle or persist/reuse the completed clone, and add a repeated-synthesis regression test. Official workflow: https://platform.minimax.io/docs/api-reference/voice-cloning-clone
2. The clone ID currently falls back to the preset voice `English_expressive_narrator`, and `sanitizeVoiceId()` does not enforce all provider rules. The UI needs a dedicated clone ID and validation: 8–256 chars, starts with an English letter, only letters/digits/`-`/`_`, cannot end in `-` or `_`, and must be unique.
3. This PR removes `voiceDesign` and `voiceDesignDesc` from `mimoOptionsFor()`. That is an unrelated MiMo regression; please restore the existing voice-design behavior.
4. MiniMax currently reuses `mimoVoiceCloneDataUri` / filename / format state, and `openaiOptionsFor()` can attach that fallback reference audio to other server TTS providers. Please keep MiniMax clone state provider-scoped and only include it for MiniMax clone mode.
5. The new `Voice mode`, `Preset voice`, `Voice clone`, upload/clear labels and help text are hardcoded English. Please add them to every locale file.
6. This branch overlaps #2329 and conflicts in the presets/tests. Please rebase after #2329, resolve the conflicts, and expose `voiceClone` consistently for both international and China MiniMax presets.

Please also add client coverage for the configurator save flow and server coverage proving that a second TTS request reuses the clone instead of uploading/cloning again.

PR #2371, opened 2026-08-06
fix(coding-agent): coalesce streamed tool start events

EKKOLearnAI:
## Summary

- buffer `response.function_call_arguments.delta` updates without emitting repeated `tool.started` events
- emit one `tool.started` with complete arguments at the function-call completion boundary
- preserve immediate starts for native Claude Code and Codex events that already contain complete arguments
- add regression coverage for streamed arguments, duplicate completion boundaries, and already-complete calls

## Root cause

Coding Agent translated every function-call argument delta into another client-facing `tool.started` event. Single chat repeatedly updated the same tool card, while Group Chat serialized every update through message persistence, context refresh, Socket broadcasts, and ACKs. Large room histories amplified that work and made tool JSON appear one character at a time.

## Impact

Single chat and Group Chat now receive one start event per Coding Agent tool call. Group Chat avoids per-delta database transactions and Socket acknowledgements. Tool result association, persisted assistant/tool messages, reasoning boundaries, completion/failure events, and native complete-argument starts remain intact.

Tool duration for streamed calls now starts when arguments are complete, so it measures execution rather than including model argument-generation time. Hermes Agent, Ekko Agent, and the raw Responses/Anthropic adapter streams are unchanged.

## Validation

- `npm test -- --reporter=dot`
- `npx vitest run tests/server/agent-runner-utils.test.ts tests/server/response-stream-reasoning-storage.test.ts tests/server/group-chat-agent-workspace.test.ts tests/server/group-chat-history-window.test.ts`
- `npm run harness:check`
- `git diff --check`

### DIFF
--- a/docs/chat-chain-changes/2026-08-05-group-chat-coding-agent-tool-events.md
+++ b/docs/chat-chain-changes/2026-08-05-group-chat-coding-agent-tool-events.md
@@ -1,0 +1,29 @@
+---
+date: 2026-08-05
+pr: pending
+feature: Group Chat Coding Agent tool event authority
+impact: Claude Code and Codex group turns no longer persist duplicate proxy tool calls beside their authoritative native tool lifecycle, Responses-backed Claude tools keep one stable call identity with compatible optional arguments, and failed tools retain their error state.
+---
+
+While a Claude Code print child is active, its native stream-json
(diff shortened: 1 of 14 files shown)


EKKOLearnAI:
Superseded by #2372, recreated from the latest main branch so the PR contains only the tool-start coalescing change.

PR #2443, opened 2026-08-09
feat(events): durable outbound event outbox with signed webhook delivery

twuijri:
Studio currently has no way to tell another system that something happened. A run finishes, a scheduled job fails, an approval sits waiting — and anything outside Studio has to poll to find out, or miss it.

This adds a **durable outbound event outbox**: operational events are recorded in SQLite first, then delivered to webhook endpoints the user configured, with retries that survive both a receiver being down and Studio restarting.

It is vendor-neutral by construction — plain signed HTTP POSTs, no push-service or mobile-vendor code anywhere.

## Reliability

The point of an outbox rather than an inline webhook call is that an inline call is lost the moment the receiver is unreachable or the process dies mid-request.

- The event is written **before** any delivery is attempted, and the insert plus its delivery rows go in one transaction — either the event is recorded with its work queued, or nothing is.
- Nothing is published until the underlying result is persisted, so a receiver reacting to `chat.run.completed` can immediately read that run back over the API.
- `event_outbox_deliveries` is separate from `event_outbox`, so one event fans out to several endpoints and each keeps its own attempt count and status.
- On startup the dispatcher drains whatever the previous process left behind. A delivery that was claimed and never finished is reclaimed after a stale window instead of being stranded in `delivering` forever.
- Publishing never blocks or fails a run: it is a local insert wrapped so a broken outbox cannot take a chat run down with it.

Delivery rules: `2xx` is success; network errors, `408`, `429` and `5xx` retry with exponential backoff and full jitter (5s → 1h); other 4xx are permanent. Redirects are not followed, requests time out at 10s, responses are read up to 64 KB, and a delivery that exhausts `max_attempts` becomes a dead letter kept for inspection rather than retried forever.

## Integrations and automation

Eleven event types across chat, group chat, cron and workflow. This PR wires `chat.run.completed` and `chat.run.failed` and defines the rest; cron and workflow producers follow in a second PR, as suggested, to keep this one reviewable.

The envelope is versioned (`schema_version: 1`) and carries a **stable event id** derived from the dedupe key, so a retried or replayed run maps to the same id and receivers can deduplicate on `id` alone.

`GET /api/hermes/capabilities` lets a client discover whether a given Studio supports the outbox, which event types it knows, and whether any endpoint is configured — so apps can hide the feature on older instances instead of guessing.

## Privacy

The envelope carries identifiers and counts. **It never carries conversation content** — no message text, no prompt, no output. A receiver that needs the text asks the API for it with its own credentials, so a leaked webhook URL does not leak what anyone said. There is a test asserting no content-bearing field appears in a default payload.

The feature is opt-in and silent: with no endpoint configured, nothing is delivered and no request leaves the instance. Secrets are write-only over the API (`has_secret`, never the value) and can be kept out of the database entirely with `secret_env`.

Requests are signed with `HMAC-SHA256(secret, timestamp + "." + rawBody)` and carry `X-Hermes-Event`, `X-Hermes-Delivery`, `X-Hermes-Timestamp`, `X-Hermes-Signature-256`. Endpoint management is super-admin only, since it decides where operational data leaves the instance.

## Tests

20 new cases in `tests/server/event-outbox-store.test.ts` and `tests/server/event-outbox-dispatcher.test.ts`, each against a real in-memory SQLite instance:

- event insertion, and deduplication that also refuses to queue a second delivery
- fan-out to multiple endpoints; filtering by event type, by profile, and by disabled state
- successful delivery, retry after `5xx` then success, retry classification (`408`/`429`/`5xx` vs `400`)
- dead-lettering on a permanent failure and on exhausted attempts
- restart recovery — a delivery left in `delivering` by a dead process is picked up
- no double-claiming of the same delivery
- HMAC signature exactly over `timestamp + "." + rawBody`, verified and negatively verified; no signature header without a secret; `secret_env` precedence
- backoff bounded below and above, and increasing with attempts
- no conversation content in a default payload

`npm run harness:check`, the server typecheck and `openapi:generate` all pass. The full `tests/server` suite is green apart from `studio-mcp-autoinject`, which fails identically on an untouched checkout of `main` — I verified that by stashing this branch.

## Docs

`docs/event-outbox.md` covers the envelope, the signing scheme, delivery semantics, the API, and a complete Express receiver that verifies the timestamp and signature in constant time and deduplicates on `event.id`. It states plainly that the feature sends nothing until the user configures an endpoint.


EKKOLearnAI:
Thanks for working on this. We now have [#2452](https://github.com/EKKOLearnAI/hermes-studio/pull/2452), which implements the Chat Run webhook integration we plan to use going forward.

It covers the full runtime lifecycle across direct chat, coding agents, workflows, Global Agent, and Group Chat: user message creation, queued/started runs, tool start/completion/failure, approval and clarification request/resolution, and run completion/failure. It also includes the super-admin settings UI, all supported locales, separate content privacy controls, SSRF/HMAC protections, and a built-in loopback receiver with a local event inbox for testing without another service.

Please use #2452 for this feature.

twuijri:
Makes sense — #2452 covers considerably more ground than this, and the feature clearly belongs with the direction you already have planned. Closing this one.

Thanks for saying so directly rather than leaving it open; that saved me from building on top of it.

I'll keep to smaller, self-contained fixes from here — happy to help with this area as a reviewer or with follow-ups on #2452 if that's ever useful.

Issue #1848, opened 2026-06-30
Bug: resolveHermesPath rejects all absolute paths via startsWith("/") check


jf7642132:
## Bug Description

In `packages/server/src/services/hermes/file-provider.ts`, the `resolveHermesPath` function contains an overly strict path validation check that blocks ALL absolute paths, including legitimate local filesystem paths.

## Problematic Code

```typescript
// file-provider.ts, resolveHermesPath()
if (normalized.startsWith("..") || normalized.includes("/../") || normalized.startsWith("/")) {
    throw Object.assign(new Error("Invalid file path"), { code: "invalid_path" })
}
```

The `normalized.startsWith("/")` check rejects every absolute path. This breaks the file browser when it requests paths like `/root/.hermes` or even the root directory itself.

## Root Cause

The path traversal protection is already handled by the `isPathWithin(resolved, homeDir)` check at the end of the function. The `startsWith("/")` guard is redundant and harmful — it prevents legitimate absolute path access while the containment check already prevents directory traversal.

## Secondary Issue

The `normalized.startsWith("..")` check is also imprecise — it would reject legitimate filenames like `..hidden` or `...`. It should be `(normalized === ".." || normalized.startsWith("../"))`.

## Proposed Fix

```typescript
// Replace:
if (normalized.startsWith("..") || normalized.includes("/../") || normalized.startsWith("/")) {
// With:
if ((normalized === ".." || normalized.startsWith("../")) || normalized.includes("/../")) {
```

## Impact

- WebUI file browser cannot list root directory or any absolute path
- Affects all 0.6.x versions of hermes-web-ui
- Users must navigate to subdirectories first as a workaround

## Environment

- hermes-web-ui: 0.6.22
- Hermes Agent: v0.17+
- Platform: Linux (also affects macOS/Windows)

PR #2446, opened 2026-08-09
fix: use SAVEPOINT for nested SQLite transactions and fix bridge broker race condition

waterWang:
## Summary

Two fixes for Hermes Studio stability issues that cause process crashes and bridge broker unavailability:

### 1. SQLite nested transaction crash (Refs #2412)

`withImmediateTransaction` in group-chat storage used `BEGIN IMMEDIATE` which cannot be nested. When `pruneMessages` was called inside `saveMessageAndRefreshRoom`'s transaction, SQLite threw `cannot start a transaction within a transaction`, an uncaught exception that killed the entire server.

**Fix:** Use SQLite `SAVEPOINT` for nested invocations. The outer call still uses `BEGIN IMMEDIATE`; any nested call creates a named savepoint instead. Rollback only reverts to the savepoint and releases it, preserving the outer transaction.

### 2. Bridge broker `_wait_ready` race condition (Refs #1965)

`WorkerProcess._wait_ready()` checked `proc.poll()` **before** draining the stdout line queue. A worker that printed `{"event": "ready"}` and then exited (e.g., watchdog timer) would trigger `"exited before ready"` before the broker ever read the `"ready"` line from the queue.

**Fix:** Swap the order — read the queue first, then check `proc.poll()`. If the worker had already signaled readiness, the queue read returns it before the poll check fires.

## Validation

- `npm run test:coverage` — 446 passed, 3 failed (timeout-only, unrelated to these changes)
- `npm run build` — builds clean
- Group-chat baseline tests: 21/21 passed
- Group-chat history window tests: 10/10 passed

## Files changed

- `packages/server/src/services/hermes/group-chat/index.ts` — SAVEPOINT fix
- `packages/server/src/services/hermes/agent-bridge/python/bridge_transport.py` — `_wait_ready` race fix

### DIFF
--- a/packages/server/src/services/hermes/agent-bridge/python/bridge_transport.py
+++ b/packages/server/src/services/hermes/agent-bridge/python/bridge_transport.py
@@ -117,11 +117,11 @@ class WorkerProcess:
         threading.Thread(target=read_stdout, daemon=True, name=f"hermes-bridge-worker-stdout-{self.key}").start()
         deadline = time.time() + self.STARTUP_TIMEOUT_SECONDS
         while time.time() < deadline:
-            if proc.poll() is not None:
-                raise RuntimeError(f"profile worker {self.key} exited before ready")
             try:
                 line = lines.get(timeout=0.1)
             except queue.Empty:
+                if proc.poll() is not None:
+                    raise RuntimeError(f"profile worker {self.key} exited before ready")
                 continue
             if line is None:
                 time.sleep(0.05)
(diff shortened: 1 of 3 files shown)


waterWang:
**Added fix for #1848 — resolveHermesPath now allows absolute paths**

The `normalized.startsWith('/')` guard was redundant: `isPathWithin(resolved, homeDir)` at the end of `resolveHermesPath` already prevents directory traversal. The extra guard blocked all absolute paths (e.g., `/root/.hermes`, `/tmp`), breaking the file browser.

**Validation:** `npm run build` passes. Local install tested — Studio serving at port 8648, HTTP 200.

PR #2450, opened 2026-08-09
fix(group-chat): incremental room token accounting instead of full-window re-tokenization (#2444)

### LINKED ISSUE #2444: fix(group-chat): prevent token accounting from starving the server event loop
## Bug

Under sustained Group Chat activity, the server can spend nearly all main-thread CPU synchronously recomputing room token totals after every persisted message. New HTTP requests, including the lightweight `/livez` endpoint, can then be delayed long enough for the runtime health check to fail.

## Root cause

`saveMessageAndRefreshRoom()` persists one message and then reloads the recent room context and runs tokenizer encoding over the entire window. Repeating that work for every tool/assistant message makes the cost proportional to the accumulated room history for every write. A single large persisted tool result is repeatedly tokenized on all subsequent messages.

A production CPU profile from an affected runtime attributed more than 94% of sampled CPU to the call chain:

`GroupChatServer.handleMessage -> ChatStorage.saveMessageAndRefreshRoom -> estimateRoomTotalTokens -> tokenizer encode`

No private endpoints, room IDs, provider names, or user content are included here.

## Expected behavior

- Persisting one Group Chat message must not synchronously retokenize the complete room history.
- Room token totals must remain correct for inserts, updates, context-window eviction, restart, and room clearing.
- Large room history and concurrent Agent messages must not starve `/livez` or unrelated HTTP traffic.
- Any full rebuild required for legacy data must be bounded/yielding and must not execute as an unbounded synchronous Socket.IO handler.

## Acceptance

- RED regression reproduces repeated full-history tokenization on the current implementation.
- The fixed path performs bounded incremental work per persisted message.
- Existing Group Chat history/window/security tests remain green.
- A stress test with large tool output, a populated room, and concurrent message writes keeps `/livez` responsive within the health-check budget.

suantea:
## What

Fixes #2444 — `saveMessageAndRefreshRoom` re-ran the tokenizer over the whole `GROUP_CHAT_MESSAGE_WINDOW` on every persisted message. Under sustained group-chat activity that is O(window) synchronous work per write; the production profile attributed 94%+ of sampled CPU to `handleMessage → saveMessageAndRefreshRoom → estimateRoomTotalTokens`, starving the event loop and tripping the `/livez` health check.

## Change

`ChatStorage` now keeps a per-room running token total (in-memory, per-server-process):

- `estimateMessageTokens(msg)` — ONE message's token contribution, mirroring the per-message logic of `estimateUsageTokensFromMessages` (content + tool_calls + reasoning, user vs assistant/tool split).
- `seedRoomTokenCache(roomId)` — one bounded full pass on a cold cache (server start / first message in a room).
- `applyMessageTokenDelta(roomId, stored, existing)` — O(1) on a warm cache: add or update that message's contribution, FIFO-evict past the window, mirroring `getMessagesForContext(excludeWorkspaceDiff: true)`. Workspace-diff rows stay excluded, same as before.
- `saveMessageAndRefreshRoom` and `saveWorkspaceDiffMessageForRun` now use the incremental path; `clearRoomContext` / `deleteRoom` invalidate the cache.

## Tests

- Regression test in `group-chat-history-window.test.ts`: after seeding 200 messages, saving one more now tokenizes only that message (< 10 `countTokens` calls) instead of the whole window (~200 calls before the fix).
- Existing suites stay green: history-window 18/18, baseline 21/21, member-sync, workspace, workspace-diff — 82/82 total.

Refs #2444

### DIFF
--- a/packages/server/src/services/hermes/group-chat/index.ts
+++ b/packages/server/src/services/hermes/group-chat/index.ts
@@ -957,10 +957,11 @@ class ChatStorage {
         if (!entry) return this.seedRoomTokenCache(roomId)
         if (stored.tool_name === 'workspace_diff') return entry.total
         const tokens = this.estimateMessageTokens(stored)
-        const prev = existing ? entry.contributions.get(existing.id) : undefined
-        if (prev !== undefined) {
+        const prevId = existing?.id
+        const prev = prevId ? entry.contributions.get(prevId) : undefined
+        if (prev !== undefined && prevId) {
             entry.total += tokens - prev
-            entry.contributions.set(existing.id, tokens)
+            entry.contributions.set(prevId, tokens)
         } else {
             entry.total += tokens
             entry.contributions.set(stored.id, tokens)
--- a/packages/server/src/services/hermes/group-chat/index.ts
+++ b/packages/server/src/services/hermes/group-chat/index.ts
@@ -898,6 +898,87 @@ class ChatStorage {
         return usage.inputTokens + usage.outputTokens
     }
 
+    // ─── Incremental token accounting (#2444) ────────────────────────────────
+    //
+    // saveMessageAndRefreshRoom used to reload the whole GROUP_CHAT_MESSAGE_WINDOW
+    // and re-run the tokenizer over it on EVERY persisted message — O(window)
+    // synchronous work per write. Under sustained group-chat activity that
+    // starved the event loop (94%+ sampled CPU in the production profile). The
+    // cache below keeps each room's running total incrementally: a save adds or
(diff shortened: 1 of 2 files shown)


EKKOLearnAI:
Thanks for working on #2444 and for adding the incremental token-accounting approach. This issue has already been addressed by #2445, which is now merged into `main` (commit `87180578`). The merged implementation also covers persisted authoritative token totals, accounting-version migration, timestamp-boundary handling, message updates/moves, and the related regression tests. Since this PR now overlaps with and conflicts with the fix already on `main`, we are closing it as superseded by #2445. Thanks again for the contribution!

PR #2451, opened 2026-08-09
feat(group-chat): make agent turn timeout configurable (#2386)

### LINKED ISSUE #2386: [Feature]: Make group chat agent turn timeout configurable (still hardcoded 120s in Hermes bridge path)
## Problem

Group Chat agent turns still hit a **hardcoded 120-second deadline** in the Hermes bridge path, and the value is not configurable.

We run hermes-web-ui 0.6.37 with a Codex participant (`@kaifazhe`) in a group chat. A typical real task (read progress doc → fix bugs → run tests → git commit → update doc) reliably dies at ~120s with:

```
Error: chat-run timed out after 120000ms
```

The task itself keeps running fine — the group chat turn just gets killed by the caller-level deadline. In our case the doc was fully written; only the final `git commit` was cut off.

## Current state

- #2356 removed the fixed `timeoutMs: 120000` from the **non-Hermes** `chatRunService.runAndWait()` call, and explicitly scoped it as "Do not change Hermes Agent Bridge". Merged in 0.6.38/0.6.39.
- In 0.6.37 the caller-level `timeoutMs: 12e4` is still passed to `runAndWait()`.
- The **Hermes participant** path (`replyToMention` → `bridge.streamOutput(run_id, {timeoutMs: 12e4})`) still hardcodes 120s in the current release — this was explicitly out of scope of #2356.
- #2385 is a related but different issue (MCP tool layer 300s default timeout for `hermes_studio_use_chat_run`), not the group-chat turn deadline.

## Request

Make the group chat agent turn timeout **configurable** instead of hardcoded:

1. Room-level or global setting (e.g. per-agent or per-room `turn_timeout_ms`), with a sane default (120s is too short for coding-agent tasks that read code, run tests, and commit).
2. Apply it consistently to both paths: `chatRunService.runAndWait()` (coding agents) and the Hermes bridge `streamOutput()` (Hermes participant).
3. Keep an upper bound to avoid unbounded runs, or route long turns through the existing background-delegation mechanism.

## Workaround we applied locally

Patched `dist/server/index.js` in 0.6.37: `timeoutMs: 12e4` → `timeoutMs: 6e5` (10 min) for the two group-chat call sites (`runAndWait` caller options + `streamOutput`). Works, but is overwritten on every `npm update` — hence the request for an official config option.

## Environment

- hermes-web-ui 0.6.37 (npm global install, Debian 13, systemd)
- Group chat with Codex participant (DeepSeek provider), long multi-step coding tasks

suantea:
## What

Fixes #2386 — the Hermes bridge path hardcoded a 120-second deadline on every group-chat agent turn. Real tasks (read → fix → test → commit → docs) reliably died at ~120s with `Error: chat-run timed out after 120000ms` even though the task itself kept running fine.

## Change

- `agent-clients.ts`: new exported `groupChatAgentTurnTimeoutMs()` reads `HERMES_GROUP_CHAT_AGENT_TURN_TIMEOUT_MS` (ms). Default `120000`; unparseable, zero, or negative values fall back to the default; fractional values are floored. Mirrors the existing `HERMES_GROUP_CHAT_MAX_AGENT_MENTION_DEPTH` env convention.
- The group-chat bridge `streamOutput` deadline now uses that value instead of the literal `120000`.

## Tests

New `group-chat-agent-turn-timeout.test.ts` (4 cases): default when unset, custom env value, fallback for unparseable/zero/negative/empty, and flooring of fractional values.

- Unit test: 4/4 green.
- `tsc --noEmit -p packages/server/tsconfig.json`: clean.

Refs #2386

### DIFF
--- a/docs/chat-chain-changes/2026-08-09-pr2451-group-chat-agent-turn-timeout.md
+++ b/docs/chat-chain-changes/2026-08-09-pr2451-group-chat-agent-turn-timeout.md
@@ -1,0 +1,18 @@
+---
+date: 2026-08-09
+pr: 2451
+feature: Group Chat agent turn timeout configurable
+impact: Room-agent Hermes turns now use a configurable timeout instead of the hardcoded 120s, so long real tasks no longer die at the 120s ceiling.
+---
+
+Group Chat room-agent turns previously ran with a hardcoded 120000ms
+deadline in the Agent Bridge streaming path (`streamOutput` timeout), which
+killed long real tasks (read → fix → test → commit → docs) with
+"chat-run timed out after 120000ms" (#2386).
+
+The timeout is now read from `HERMES_GROUP_CHAT_AGENT_TURN_TIMEOUT_MS`
+(defaulting to 120000 when unset or invalid), mirroring the existing
+`HERMES_GROUP_CHAT_MAX_AGENT_MENTION_DEPTH` env-based configuration in the
+group-chat runtime. The stream wait between starting a run and reading the
+first chunk uses the configured value, so deployments with longer-running
+agents can raise the ceiling without a code change.
--- a/packages/server/src/services/hermes/group-chat/agent-clients.ts
+++ b/packages/server/src/services/hermes/group-chat/agent-clients.ts
@@ -24,6 +24,16 @@ import { buildAgentInstructions, buildNonOwnerRequestSecurityPrompt } from '../c
 
 export const GROUP_CHAT_AGENT_SOCKET_SECRET = randomBytes(32).toString('hex')
 
+// Group-chat Hermes turns used a hardcoded 120s deadline in the bridge path;
+// long real tasks (read → fix → test → commit → docs) reliably died at 120s
+// with "chat-run timed out after 120000ms" (#2386). Configurable via env,
+// mirroring HERMES_GROUP_CHAT_MAX_AGENT_MENTION_DEPTH in group-chat/index.ts.
+export function groupChatAgentTurnTimeoutMs(): number {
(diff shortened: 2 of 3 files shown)


EKKOLearnAI:
Thanks for looking into #2386. We re-checked the current `main` execution paths and this PR is changing a Bridge request timeout, not a Group Chat turn timeout:

- The non-Hermes Group Chat path no longer passes a 120-second timeout to `runAndWait` (removed by #2357), so Codex, Claude, and Ekko turns do not have that absolute deadline.
- The Hermes path starts the Agent run on a background Python thread. `streamOutput(..., { timeoutMs: 120000 })` applies to each individual `get_output` IPC request; it is not a 120-second deadline for the full turn, and the polling timer resets on every request.
- Generic Agent Bridge request timeouts are already configurable through `HERMES_AGENT_BRIDGE_TIMEOUT_MS`.
- Group Chat rolling context summaries currently use the bare Ekko summarizer with `timeoutMs: 300_000`, so compression already has a five-minute model-request budget.

Because of this, `HERMES_GROUP_CHAT_AGENT_TURN_TIMEOUT_MS` and the PR description mischaracterize the existing behavior, and the change does not address an active full-turn timeout on current `main`. I recommend not merging this as-is. If there is still a reproducible timeout on current `main`, please include the exact error text and call path so we can adjust the actual deadline involved.

EKKOLearnAI:
Follow-up: I opened #2457 for the separate concern we found during review. It makes the production Group Chat rolling-summary compression timeout configurable while preserving the existing five-minute default. It does not add or change any Agent turn deadline.

PR #2457, opened 2026-08-10
feat(group-chat): configure compression timeout

EKKOLearnAI:
## What

- Add HERMES_GROUP_CHAT_COMPRESSION_TIMEOUT_MS for the production Ekko rolling-summary request.
- Keep the existing 300000ms default.
- Accept bounded overrides from 5000ms through 1800000ms and fall back safely for invalid values.
- Add regression coverage for parsing and the actual Ekko provider timeout wiring.

## Why

Group Chat rolling summaries already have a five-minute model-request budget, but the value was hardcoded. Deployments with different providers and history sizes could not tune compression independently. This setting is intentionally separate from Agent turn limits and Agent Bridge per-request timeouts.

## Impact

Default behavior does not change. Operators can tune only the rolling-summary compression request without introducing an absolute deadline for Hermes, Ekko, Codex, or Claude turns.

## Checks

- npm run test -- tests/server/group-chat-summary-timeout.test.ts tests/server/group-chat-room-summary.test.ts
- npm run harness:check
- npx tsc --noEmit -p packages/server/tsconfig.json


EKKOLearnAI:
Closing this draft after clarifying the intended behavior: Group Chat rolling summaries should keep the existing fixed 300-second Ekko model-request timeout and do not need a separate environment override. Current main already uses timeoutMs: 300_000 on the production GroupRoomSummaryService path, so no code change is required.

PR #2693, opened 2026-08-23
fix(chat): count thinking time from the run, not from the page

twuijri:
The thinking indicator's elapsed time starts at the client's first render:

```ts
thinkingStartedAt = Date.now();
thinkingElapsedMs.value = 0;
```

That watcher runs with `{ immediate: true }`, so it fires when the component mounts — which is when you *open* the session, not when the run began.

Reported from a live instance: leave a session while the agent is working, come back, and it reads a few seconds again as though it had just started. With the same run open on two devices, the one left alone keeps counting correctly while the one that navigated away resets — same run, two different numbers.

## What changed

The run's start is only known server-side, so `SessionState` gains `runStartedAt`, set at both points where a run begins working, and the `resumed` payload carries it. The chat store keeps it per session and drops it when the session reports it is no longer working. The timer uses it as its origin.

`Date.now()` stays as the fallback when the field is absent, so a client talking to an older server behaves exactly as it does now rather than showing nothing.

The indicator's visibility rule, the one-second tick and the formatting are untouched.

## Tests

`tests/server/run-chat-run-started-at.test.ts` drives the socket: resuming a working session emits `resumed` carrying the recorded `runStartedAt`, and a session that is not working reports no start at all.

Reverting `run-chat/index.ts` fails the first. `vue-tsc -b`, server `tsc` and `harness:check` pass, with a chat-chain fragment included. The full `tests/client` and `tests/server` run shows only the `use-speech-tts` and `studio-mcp-autoinject` failures that reproduce on a detached `origin/main`.

### DIFF
--- a/docs/chat-chain-changes/2026-08-22-thinking-elapsed-run-start.md
+++ b/docs/chat-chain-changes/2026-08-22-thinking-elapsed-run-start.md
@@ -1,0 +1,16 @@
+---
+date: 2026-08-22
+feature: thinking-timer-counts-from-the-real-run-start
+pr: 2693
+impact: The thinking timer shows how long the agent has actually been working instead of counting from the client's first render. No change to when the indicator appears or to run handling.
+---
+
+# Thinking timer origin
+
+`MessageList` started the elapsed clock with `thinkingStartedAt = Date.now()` the moment the run indicator became visible. That is the client's first render, not the run's start, so leaving a page mid-run and coming back showed the agent as having just started, and two devices watching one run disagreed — the one that stayed open kept counting, the one that navigated reset to zero.
+
+The run's start is only known server-side, so `SessionState` now carries `runStartedAt`, set at both points where a run begins working, and it is included in the `resumed` payload. The chat store keeps it per session, clearing it when the session reports it is no longer working, and the timer uses it as its origin.
+
+`Date.now()` remains the fallback when the value is absent, so a client talking to an older server behaves exactly as it does today rather than showing nothing.
+
+Nothing else changes: the indicator's visibility rule, the one-second tick, and the formatting are untouched.
--- a/packages/client/src/api/hermes/chat.ts
+++ b/packages/client/src/api/hermes/chat.ts
@@ -162,6 +162,8 @@ export interface ResumeSessionPayload {
   hasMoreBefore?: boolean
   isWorking: boolean
   isAborting?: boolean
+  /** Epoch ms the active run began; absent on servers that predate it. */
+  runStartedAt?: number
   events: Array<{ event: string; data: RunEvent }>
   inputTokens?: number
   outputTokens?: number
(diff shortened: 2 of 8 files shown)


EKKOLearnAI:
Thanks for the fix. The server-side runStartedAt plumbing looks good, but there are a few frontend lifecycle gaps that should be addressed before merging. These should not require another server change:

1. The per-session runStartedAt value is only deleted in the switchSession resumed callback when isWorking is false. Normal completion/error/abort cleanup does not clear it. After a session has been resumed mid-run, the next new run in that same session can therefore reuse the previous run timestamp and show a much larger elapsed time.
2. MessageList watches only isRunIndicatorActive. If the user switches between two sessions that are both already known as working, the watched value stays true, so the timer is not restarted and keeps using the previous session timestamp. The watcher should also react to activeSessionId and the active session runStartedAt value.
3. The other resume paths do not apply runStartedAt: applyReconnectResume and the visibilitychange resume callback currently update isWorking but ignore the timestamp. Returning to the foreground or reconnecting can still start the clock from the client render time.

A small shared frontend helper for applying/clearing the timestamp from ResumeSessionPayload, used by every resume path and terminal cleanup path, should cover this. Please also add regression coverage for: resume mid-run then complete/start another run, switching between two working sessions, and reconnect/visibility resume.

Non-blocking design note: subtracting a server epoch timestamp from browser Date.now() assumes synchronized clocks. If exact cross-device agreement is required later, the protocol could also return serverNow or elapsedMs.

twuijri:
Thanks — all three were real, and the second one is the worst of them: two working sessions kept the same origin because the watched value never changed.

Pushed:

- `applyResumedRunStartedAt` / `clearRunStartedAt` in the store. Every resume path uses the first — `switchSession`, `applyReconnectResume`, and the `visibilitychange` resume — and every terminal path uses the second: the terminal run event, the run's own `cleanup()` on completion, error and abort, and session destroy.
- `MessageList` now watches `[isRunIndicatorActive, activeSessionId, runStartedAt.get(activeSessionId)]`, so switching between two working sessions restarts the timer on the right origin.

Regression coverage in `tests/client/chat-store-run-started-at.test.ts`: the start survives a mid-run resume; it is gone once the run completes, so the next run in that session cannot inherit it; two working sessions keep their own values; a start arriving with `isWorking: false` is ignored; and the widened watcher and the `Date.now()` fallback are pinned. Reverting the store fails the completion case.

On the clock-skew note: agreed, and `Date.now()` on both sides is what the elapsed display already assumed before this PR, so this does not make it worse. Returning `serverNow` alongside `runStartedAt` and correcting for the offset is a small follow-up if you want exact cross-device agreement — happy to do it in a separate PR rather than widen this one.

twuijri:
Closing — this landed in #2723, which rebuilt it on current main and added the tool-run grouping and clock-skew handling on top. Thanks for the review; all three gaps you found were real.

PR #2711, opened 2026-08-23
feat(chat): show time-to-first-token (TTFT) for assistant messages

monikalnbo:
## What

Adds an optional TTFT badge to assistant messages in the single-chat view, measuring the time from run issue to the first `message.delta`.

Purely **additive** — no existing code paths, payloads, or persistence are modified.

## How

- New util `packages/client/src/utils/hermes/ttft-tracker.ts`: records per-session run start time; first content delta consumes it and returns elapsed ms.
- `stores/hermes/chat.ts`: 
  - optional `ttftMs?: number` field on `Message`
  - `noteRunStart(sid)` right before `startRunViaSocket`
  - `consumeFirstDelta(sid)` in the existing `message.delta` case, attached to the streaming/new assistant message via `??=` semantics
- `MessageItem.vue`: renders a small badge in the existing `.message-meta` row **only when** a measurement exists
- i18n: `chat.ttft` / `chat.ttftTitle` added to all 11 locales

44 insertions, 0 deletions across 13 files.

## Scope notes

- Measures client-perceived latency (emit → first rendered token), which includes network + server queueing. A follow-up could thread provider-side TTFT through usage reporting if stricter numbers are wanted.
- Only the single-chat flow is instrumented; group chat / global agent can adopt the same tracker later.
- Sessions resumed without a local run start simply return null and render nothing.

## Validation

- esbuild TS transform passes for all touched files
- Full `npm ci && npm run harness:check && npm run test` not run in this environment (no dependency install); happy to run locally before un-drafting

Submitted as draft for early feedback.

### DIFF
--- a/packages/client/src/components/hermes/chat/MessageItem.vue
+++ b/packages/client/src/components/hermes/chat/MessageItem.vue
@@ -1197,6 +1197,13 @@ onBeforeUnmount(() => {
             </span>
           </div>
           <div class="message-meta">
+            <span
+              v-if="message.role === 'assistant' && message.ttftMs != null"
+              class="thinking-meta ttft-meta"
+              :title="t('chat.ttftTitle')"
+            >
+              {{ t('chat.ttft', { ms: message.ttftMs }) }}
+            </span>
             <button
               v-if="canPlaySpeech"
               class="speech-bubble-btn"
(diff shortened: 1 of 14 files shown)


monikalnbo:
Closing for now — will resubmit after syncing with latest main.

PR #2810, opened 2026-08-31
fix(group-chat): declare the conversation each reply belongs to

JoaoMarcos44:
`Refs NousResearch/hermes-agent#96811` — this is the host half of that issue.

Group chat re-keys **every** conversation-affinity hint Hermes sends, on **every single reply**, so a room member never lands back on the prefix it just warmed.

## The defect

`groupRuntimeSessionId(roomId, profile, name)` mints a fresh UUID-backed id per reply. That is correct and deliberate — it is a server-issued continuation identity, never accepted from an Agent socket. The problem is that this id is the *only* conversation signal the bridge hands Hermes, and Hermes derives all four affinity surfaces from it.

| Surface | Where it goes |
| --- | --- |
| `prompt_cache_key` | both OpenAI-wire transports |
| `body.session_id` | OpenRouter sticky routing |
| `body.session_id` | Nous Portal sticky key |
| `x-grok-conv-id` | xAI header |

Hermes deliberately refuses to recover a conversation from an id's *shape* — client-supplied ids collide, and truncated room-member ids collide with each other — so the conversation has to be **declared**.

```mermaid
%%{init: {'theme': 'dark', 'themeVariables': { 'primaryColor': '#8b0000', 'mainBkg': '#0a0204', 'primaryTextColor': '#ffccd5', 'primaryBorderColor': '#ff0038', 'lineColor': '#ff0038'}}}%%
graph TD
    A[🩸 Room Member Replies] --> B[groupRuntimeSessionId<br/>fresh UUID every reply]
    B --> C{Conversation Declared?}

    C -->|before: no| D[🔥 Hermes Sees A New<br/>Conversation Each Reply]
    D --> E1[prompt_cache_key rotates]
    D --> E2[OpenRouter sticky id rotates]
    D --> E3[Nous Portal key rotates]
    D --> E4[x-grok-conv-id rotates]
    E1 --> F[⚔️ Cold Prefix Forever]
    E2 --> F
    E3 --> F
    E4 --> F

    C -->|after: yes| G[groupBridgeSessionId<br/>room + profile + name<br/>+ runtime + sessionSeed]
    G --> H[🔒 One Affinity Identity<br/>gwk_ hash]
    H --> I[Warm Prefix Across Replies]
    G --> J[New sessionSeed<br/>= new conversation]
    J --> K[Cold Bucket On Reset]
```

## The fix

The stable value already exists. `groupBridgeSessionId(roomId, profile, name, sessionSeed, runtimeConfig)` is stable for exactly one conversation: it carries the room, profile, agent name and runtime identity, plus the room-owned `sessionSeed` — which is what `UPDATE gc_rooms SET sessionSeed = ?` rotates when a room starts a new conversation. So a real room reset lands on a cold bucket without Hermes having to observe a reset of its own. The arguments match the room socket's own calls (`sockets/group-chat.ts:3704`, `:4167`) byte for byte, so both sides name one conversation.

It travels as `gateway_session_key`, an optional field the Hermes `AIAgent` constructor already accepts, threaded through the one path the reproduction uses.

```mermaid
%%{init: {'theme': 'dark', 'themeVariables': { 'primaryColor': '#8b0000', 'mainBkg': '#0a0204', 'primaryTextColor': '#ffccd5', 'primaryBorderColor': '#ff0038', 'lineColor': '#ff0038'}}}%%
graph LR
    A[replyToMention<br/>agent-clients.ts] -->|gateway_session_key| B[bridge.chat<br/>client.ts]
    B -->|JSON payload| C[bridge_server.py<br/>forwarded only when declared]
    C --> D[AgentPool.start_chat]
    D --> E[get_or_create]
    E --> F[🔥 AIAgent gateway_session_key]
    G[Non-Hermes Agents] -.->|early return| H[replyToMentionWithChatRun<br/>untouched]
```

| File | What |
| --- | --- |
| `studio/services/group-chat/agent-clients.ts` | `groupDeclaredConversationKey()`, sent on the reply |
| `hermes/services/bridge/client.ts` | optional `gateway_session_key` in the chat options and payload |
| `hermes/services/bridge/python/bridge_server.py` | reads it, forwards it only when declared |
| `hermes/services/bridge/python/bridge_pool.py` | threads it to `AIAgent`, refreshes an idle cached session |
| `tests/server/group-chat-declared-conversation.test.ts` | new |

## Blast radius

- **Only the Hermes primary-agent reply path changes.** `replyToMention` returns early into `replyToMentionWithChatRun` for every non-hermes agent, so codex, claude, ekko and pi are untouched.
- **The field is optional everywhere.** `handle-bridge-run.ts` does not send it and is unchanged. `bridge_server` forwards the keyword *only* when a conversation is actually declared, so a pool implementation that does not know the parameter keeps working — this is what keeps `agent-bridge-python-concurrency`'s `FakePool` green without editing it. `get_or_create` and `start_chat` take the parameter last, so the existing positional call still binds.
- **A cached session reused under a rotated declaration** has its key refreshed only while idle, never mid-run.
- **A room that cannot be read** declares nothing and keeps today's behaviour exactly.

## Proof, on real values

Not fixtures. The real `groupRuntimeSessionId` and `groupBridgeSessionId` produced the ids; a real Hermes `AIAgent` on a real SQLite `state.db`, with rows written exactly the way `_prepersist_user_message` writes them, produced the wire values through the real transports and provider plugins.

**Before** — three replies, three identities, and the raw runtime id leaving the process verbatim:

```
reply gc_run_room-7c1f_default_Reviewer_e1239667…   prompt_cache_key pck_999e9dc4fb9911eb9d50cbec
                                                    openrouter/nous/x-grok-conv-id = the raw gc_run_ id
reply gc_run_room-7c1f_default_Reviewer_96786905…   prompt_cache_key pck_6e10ca6401777f7521da7016
reply gc_run_room-7c1f_default_Reviewer_5803014f…   prompt_cache_key pck_abc158d968c43fd850bc8a9d
-> 3 distinct affinity identities across 3 replies
```

**After** — the same three replies, one identity on all four surfaces:

```
reply gc_run_room-7c1f_default_Reviewer_e1239667…   prompt_cache_key      pck_daf290a8a983032b4b865bf4
                                                    openrouter_session_id gwk_39165ec23d41e21816aba30b
                                                    nous_session_id       gwk_39165ec23d41e21816aba30b
                                                    x_grok_conv_id        gwk_39165ec23d41e21816aba30b
reply gc_run_room-7c1f_default_Reviewer_96786905…   identical
reply gc_run_room-7c1f_default_Reviewer_5803014f…   identical
-> 1 distinct affinity identity across 3 replies
```

And the three properties that keep it honest: a new `sessionSeed` rotates the identity, another room member never shares it, and the raw declared key never reaches the wire — it is hashed to `gwk_<sha256[:24]>` first, because it embeds room, profile and member names and this value goes out as a provider routing key.

## Test plan

- [x] `tests/server/group-chat-declared-conversation.test.ts` — **6 passed**, new. Removing the declaration turns **3 of them red**; no existing test was modified to make them pass.
- [x] `group-chat-agent-workspace` + `group-chat-agent-routing-baseline` + `group-chat-agent-model-config` + `group-chat-agent-disconnect` + `group-chat-approval` — **86 passed**.
- [x] `agent-bridge-python-concurrency` — **43 passed**, including `forwards Agent creation policy from chat and context-estimate requests`, which drives the real `bridge_server.handle` against a pool that does not know the new parameter.
- [x] `agent-bridge-client-*` and `agent-bridge-reasoning-effort` — green.
- [x] `tsc --noEmit -p packages/server/tsconfig.json` — clean.
- [x] `agent-bridge-manager` has 3 failures on this machine; they are **pre-existing** and reproduce identically on a clean `main` checkout (verified with `git stash`).

### DIFF
--- a/packages/server/src/modules/hermes/services/bridge/client.ts
+++ b/packages/server/src/modules/hermes/services/bridge/client.ts
@@ -56,6 +56,21 @@ export interface AgentBridgeChatOptions {
   /** Local patch (reasoning-effort): per-session reasoning effort override.
    * Empty/undefined = use config.yaml default. */
   reasoning_effort?: string
+  /** The conversation this reply belongs to, declared to Hermes.
+   *
+   * `session_id` names one runtime execution; a caller that issues a fresh
+   * one per reply (group chat does) leaves Hermes with no way to tell that
+   * two replies continue the same conversation, so every affinity hint it
+   * derives from that id — the OpenAI-wire `prompt_cache_key`, OpenRouter's
+   * and Nous' sticky `session_id`, xAI's `x-grok-conv-id` — is re-keyed on
+   * every single reply and never lands back on a warm prefix
+   * (NousResearch/hermes-agent#96811).
+   *
+   * Hermes deliberately will not infer the conversation from the id's shape,
+   * so it has to be declared. Pass a value that is stable for one
+   * conversation and distinct across conversations; omit it to keep the
+   * previous per-reply behaviour exactly. */
+  gateway_session_key?: string
 }
 
 export type AgentBridgeMessage =
@@ -526,6 +541,9 @@ export class AgentBridgeClient {
         : {}),
       // Local patch (reasoning-effort): per-session reasoning effort override.
       ...(options.reasoning_effort ? { reasoning_effort: options.reasoning_effort } : {}),
+      ...(options.gateway_session_key
+        ? { gateway_session_key: options.gateway_session_key }
+        : {}),
     })
   }
 
--- a/packages/server/src/modules/hermes/services/bridge/python/bridge_pool.py
+++ b/packages/server/src/modules/hermes/services/bridge/python/bridge_pool.py
@@ -384,6 +384,35 @@ class AgentPool:
             except Exception:
                 pass
 
+    @staticmethod
+    def _apply_declared_conversation(
+        session: AgentSession, gateway_session_key: str | None
+    ) -> None:
+        """Keep a cached session's declared conversation current.
+
+        The pool is keyed by ``session_id``, which names one runtime execution.
+        The declared conversation is a different identity and can rotate under a
+        reused id — group chat rotates it whenever the room's ``sessionSeed``
+        changes. Applied only while the session is idle, so a live turn's
+        affinity cannot move under it, and never cleared by a caller that simply
+        declares nothing.
+        """
+        key = str(gateway_session_key or "").strip()
+        if not key or session.running:
+            return
+        if str(session.config.get("gateway_session_key") or "") == key:
+            return
+        session.config["gateway_session_key"] = key
+        agent = getattr(session, "agent", None)
+        if agent is not None:
+            try:
+                agent._gateway_session_key = key
+            except Exception:
+                # A runtime that will not accept the attribute keeps the key it
+                # was built with; the declaration is an optimization, never a
+                # correctness requirement.
+                pass
+
     def get_or_create(
         self,
         session_id: str,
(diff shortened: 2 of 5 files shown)


JoaoMarcos44:
Closing at the author's direction: this work is being kept inside NousResearch/hermes-agent only, and should not land in this repository. Reopening is a maintainer's call if the host adoption is ever wanted here.

PR #2816, opened 2026-08-31
fix(skills): restore per-skill toggle hidden by Vue Boolean prop casting / 修复技能页单技能开关被 Boolean 默认值吞掉的问题

WenhuaXia:
## Summary / 摘要

**EN**: Since #2770 (v0.7.0), the per-skill `NSwitch` on the Skills page is gone on both `/hermes/skills` and `/ekko/skills`. #2770 added a `toggleable?: boolean` prop to `SkillList` and gated the switch on `!readonly && toggleable !== false` — intent being "enabled unless explicitly disabled". But Vue 3's boolean casting makes an omitted `type: Boolean` prop default to `false`, so `toggleable !== false` is *always* false and the switch never renders. Neither call site passes `toggleable`.

Fix: pin the default with `withDefaults(..., { toggleable: () => true })`.

**中文**: 自 #2770（v0.7.0）起，技能页（`/hermes/skills` 和 `/ekko/skills`）每行技能的单个开关消失了。#2770 给 SkillList 加了 `toggleable?: boolean` prop，并把 NSwitch 显示条件改成 `!readonly && toggleable !== false`，本意是"默认开、显式传 false 才关"。但 Vue 3 的 boolean casting 会把未传的 Boolean prop 默认为 `false`，导致 `toggleable !== false` 恒为 false，开关永远不渲染；两个调用方都没有传 `toggleable`。修复：用 `withDefaults` 把默认值钉回 `true`。

## Runtime evidence / 运行时证据

Captured from the running Vue app (instance normalized props vs raw vnode props):

```json
{
  "instProps":     { "readonly": false, "toggleable": false, "toggleableType": "boolean" },
  "rawVnodeProps": ["categories", "archived", "selected-skill", "search-query", "source-filter", "readonly", "onSelect", "onDeleted"]
}
```

- Raw vnode props: call site does **not** pass `toggleable`
- Instance props: normalized to `false` → `toggleable !== false` = false → NSwitch renders a Vue empty comment (`<!---->`)
- Delete button (gated only on `!readonly`) still renders, confirming `readonly=false` and isolating the cause to the `toggleable` condition
- Before fix: 234 skill rows, 0 `.n-switch` elements. After fix: per-skill switch renders.

## Verification / 验证

- `npm run build`（vue-tsc + vite + server tsc）passes
- Local runtime: after the fix the per-skill NSwitch renders on both `/hermes/skills` and `/ekko/skills`; toggle round-trips through the existing `PUT /api/hermes/skills/toggle` (hermes) and `toggleHandler` / `PUT /api/ekko/skills/:name/toggle` (ekko)
- 1-line behavioral change (`withDefaults` default only); no server/API changes, no new deps

## Files / 文件

- `packages/client/src/components/hermes/skills/SkillList.vue` (+8/-2)

### DIFF
--- a/packages/client/src/components/hermes/skills/SkillList.vue
+++ b/packages/client/src/components/hermes/skills/SkillList.vue
@@ -11,7 +11,7 @@ const { t } = useI18n()
 const message = useMessage()
 const dialog = useDialog()
 
-const props = defineProps<{
+const props = withDefaults(defineProps<{
     categories: SkillCategory[]
     archived: SkillInfo[]
     selectedSkill: string | null
@@ -21,7 +21,9 @@ const props = defineProps<{
     toggleable?: boolean
     toggleHandler?: (category: string, skill: string, enabled: boolean) => Promise<void>
     deleteHandler?: (category: string, skill: string) => Promise<void>
-}>()
+}>(), {
+    toggleable: true,
+})
 
 const emit = defineEmits<{
     select: [category: string, skill: string]
--- a/tests/client/skill-list.test.ts
+++ b/tests/client/skill-list.test.ts
@@ -3,6 +3,7 @@ import { describe, expect, it, vi } from 'vitest'
 import { mount } from '@vue/test-utils'
 import { defineComponent } from 'vue'
 import SkillList from '@/components/hermes/skills/SkillList.vue'
+import { toggleSkill } from '@/api/hermes/skills'
 
 vi.mock('vue-i18n', () => ({
   useI18n: () => ({ t: (key: string) => key }),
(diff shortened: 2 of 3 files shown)


Afang17:
Same issue here — the per-skill toggle is missing on both `/hermes/skills` and `/ekko/skills` since v0.7.0. We applied the same fix locally (`toggleable: { type: Boolean, default: true }`) and verified it works: all toggles render again and the API call goes through. Also worth noting: `readonly` is evaluated at the `SkillList` component level (based on the currently selected skill's source), so the switch also disappears for all rows when a builtin skill is selected — you may want to decouple the switch from `readonly` too (readonly should still guard editing/deleting). Looking forward to this being merged. 👍

同款问题，本地按同样方案修复并验证有效。另外发现 `readonly` 是组件级判断（取决于当前选中 skill 的 source），选中内置 skill 时整列表开关也会消失，建议开关与 readonly 解耦（readonly 仍保护编辑/删除）。期待合入。

PR #2844, opened 2026-09-02
fix(desktop): keep the desktop window from stealing focus on startup

### LINKED ISSUE #2835: [Bug]: 界面无操作自弹出
### Hermes Web UI Version

v0.7.15

### Hermes Agent Version

v0.20.6

### Bug Description

hermes studio开启后缩小或聚焦在其他窗口时，web ui会反复弹出到最上层

### Steps to Reproduce

1.打开hermes studio
2.聚焦其他窗口操作时

### Expected Behavior

studio在后台休眠

### Actual Behavior

studio无操作反复弹出

### Logs / Error Messages

```shell

```

### Environment

Windows

### Node Version

_No response_

### Additional Context

_No response_

lorenzozanee:
## Summary
- Keep the desktop window visible on startup without stealing focus on Windows.
- Preserve explicit foreground activation when the user clicks tray, notifications, or second-instance entry points.

Closes #2835

## Validation
- `npm test -- tests/desktop/windows-main-window.test.ts tests/desktop/chat-window.test.ts`
- `npm run harness:check`
- `npm run build`

### DIFF
--- a/packages/desktop/src/main/index.ts
+++ b/packages/desktop/src/main/index.ts
@@ -107,9 +107,10 @@ function showWindowWithFade(focus = true) {
   if (mainWindow.isMinimized()) mainWindow.restore()
 
   cancelWindowFade()
+  const showWindow = focus ? () => mainWindow.show() : () => mainWindow.showInactive()
   if (process.platform !== 'win32' || mainWindow.isVisible()) {
     mainWindow.setOpacity(1)
-    mainWindow.show()
+    showWindow()
     if (focus) mainWindow.focus()
     return
   }
@@ -117,7 +118,7 @@ function showWindowWithFade(focus = true) {
   const durationMs = 180
   const startedAt = Date.now()
   mainWindow.setOpacity(0)
-  mainWindow.show()
+  showWindow()
   if (focus) mainWindow.focus()
   windowFadeTimer = setInterval(() => {
     if (!mainWindow || mainWindow.isDestroyed()) {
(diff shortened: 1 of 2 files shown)


Undertone0809:
Thanks for putting together this focused fix. I reviewed the patch against the newer reproduction details in #2835 and #2850.

The current change looks useful for the automatic startup path:

- `ready-to-show` changes from `showWindowWithFade(true)` to `showWindowWithFade(false)`
- explicit user actions such as tray / notification / second-instance still retain foreground activation
- CI is green

There now seem to be two distinct scopes:

1. **Startup focus stealing** through `ready-to-show` — covered by this PR.
2. **Runtime focus-stealing loops** after Studio has already been running — reported on Windows in #2850 and on macOS in #2835.

To keep this PR small and easier to merge, would it make sense to treat it explicitly as a startup-only fix, change `Closes #2835` to `Refs #2835`, and track the runtime path separately? That avoids closing the broader issue before the long-running/macOS behavior is verified.

One small testing concern: the changed helper is platform-neutral, while the regression assertion currently lives in `windows-main-window.test.ts` and checks source text. A platform-neutral focus-policy/helper test would make the intended boundary clearer, but I would not block this narrow startup fix on a larger refactor.

I can help validate a rebased candidate/package on macOS with Stage Manager enabled and disabled, using a matrix that distinguishes startup, background runtime, tray click, notification click, and second-instance activation. My preference would be to merge the narrow startup fix first once its scope is explicit, then follow up on the runtime loop with separate evidence.

PR #2928, opened 2026-09-06
feat(desktop): configure Markdown link opening target

Undertone0809:
## Summary

- Add a desktop-only **Settings → Display → Web link opening target** preference.
- Let users choose between opening ordinary Markdown HTTP(S) links in the embedded Hermes Studio browser or the OS default browser.
- Persist the choice locally, keep Web UI behavior unchanged, and leave HTML artifact previews unaffected.
- Add coverage for persistence, all locales, Markdown routing, Electron origin validation, and the desktop/Web UI split.
- Include the reviewed UI screenshots in this PR under `docs/screenshots/link-opening-preference/`.

## Review fix

- Route explicit `default-browser` requests through a dedicated `hermes-desktop:open-external-url` IPC method instead of `window.open`.
- Validate the IPC caller against a known Hermes desktop window and accept only absolute `http:`/`https:` URLs before calling `shell.openExternal`.
- Add regression coverage proving a same-origin Markdown link uses the system-browser IPC path and does not open an Electron internal child window.
- Surface system-browser IPC failures instead of falling back to `window.open`, which could re-enter the internal-window path for same-origin URLs.

### Default: Hermes Studio

Default Hermes Studio link target

### Available targets

Link opening target options

### Saved: default browser

Default browser link target saved

## Validation

- `NODE_ENV=test npm run test -- tests/client/desktop-browser-open.test.ts tests/client/display-settings.test.ts tests/client/link-open-target-locales.test.ts tests/client/markdown-rendering.test.ts tests/desktop/window-open-policy.test.ts tests/desktop/preload-auth.test.ts` — **74 tests passed**
- `PLAYWRIGHT_PORT=4201 PLAYWRIGHT_CHANNEL=chrome npx playwright test tests/e2e/link-opening-preference.spec.ts --workers=1` — **2 passed**
- `npm --prefix packages/desktop run build:main` — passed
- `npm run build` — passed

## Notes

- The preference is stored under `hermes_link_open_target` and defaults to the existing embedded-browser behavior.
- Electron validates the exact active Web UI origin before allowing an internal child window; external URLs continue to use `shell.openExternal`.
- Detached chat windows do not expose the embedded browser bridge, so links from those windows continue to use the system browser.


packages/desktop/src/main/index.ts
```diff
@@ -551,7 +552,7 @@ async function createWindow(): Promise<void> {
   mainWindow.webContents.setWindowOpenHandler(({ url, frameName }) => {
     const agentLinkPopup = groupChatAgentLinkPopupResponse(url, frameName)
     if (agentLinkPopup) return agentLinkPopup
-    if (url.startsWith('http://127.0.0.1') || url.startsWith('http://localhost')) {
+    if (isTrustedDesktopAppUrl(url, serverUrl)) {
```
EKKOLearnAI, line 556: **[P2] Honor the default-browser preference for same-origin Markdown links**

When the user selects **Default browser**, `openUrlInDesktopBrowser()` returns `false`, so `MarkdownRenderer` falls back to `window.open(url, '_blank', 'noopener,noreferrer')`. This handler still allows every URL matching the active Web UI origin, so an ordinary Markdown HTTP(S) link on that origin opens an Electron child window instead of the selected OS browser.

For example, with Studio running at `http://127.0.0.1:8748`, selecting Default browser and clicking `[Docs](http://127.0.0.1:8748/docs)` takes the `action: 'allow'` branch. I checked the actual handler in isolation: this URL does not invoke `shell.openExternal`, while external-origin URLs do.

Could you route explicit system-browser requests through a dedicated, validated desktop IPC method, or otherwise distinguish them from intentional internal-window requests? Please add coverage for a same-origin Markdown link with Default browser selected; the current renderer tests stop at `window.open`, and the policy test independently expects same-origin URLs to be allowed.

Validation on this commit: all 64 focused unit tests, both preference E2E tests, `harness:check`, the desktop main-process build, and the full build passed.

Undertone0809: Fixed in 8c6f796b. When `default-browser` is selected, ordinary Markdown HTTP(S) links now use the dedicated `hermes-desktop:open-external-url` IPC method, so same-origin URLs no longer enter the internal `setWindowOpenHandler` allow branch. The main process validates the sender against a known Hermes desktop window and accepts only absolute HTTP(S) URLs before calling `shell.openExternal`. Added a regression test covering a same-origin Markdown link and asserting the IPC path is used without `window.open`; the affected focused suite now passes 72 tests.

Undertone0809: Follow-up in 6584c1d1: if the current desktop IPC exists but `shell.openExternal` reports failure, the client now surfaces an error instead of falling back to `window.open`. That prevents a same-origin link from re-entering Electron’s internal-window path after an explicit Default browser request. Added regression coverage for both the bridge failure and the Markdown click path; the focused suite now passes 74 tests.

PR #3101, opened 2026-09-18
fix(push): retry transient iOS delivery failures

wtj-0527:
## Summary

- retry transient iOS `/push/v1/send` delivery failures (network errors, HTTP 408/429/5xx)
- avoid marking an event permanently delivered until the provider accepts it
- keep the same `event_id` across bounded retry attempts
- add safe delivery diagnostics without logging credentials or APNs tokens

## Behavior

- Android continues to use the existing WebSocket path
- iOS continues to use POST `/push/v1/send`
- both paths remain behind the shared `businessEvents.publish` event outlet
- permanent 4xx responses such as 401/403 are not retried

## Verification

- `vitest run tests/server/run-push-consumer.test.ts tests/server/chat-webhooks.test.ts` — 21 passed
- `npm run build` — passed
- `git diff --check github/main...HEAD` — passed


wtj-0527:
Closing with the author's authorization: this implementation targets the superseded per-run push_snapshot/run_push_targets consumer. Current upstream main (checked at fa131eeb46c14ce6201a5c39100087b52fef3f59) selects recipients from Studio-user-scoped user_push_devices with current connection, ownership, permission and per-device push controls, following #3098/#3102.

This PR should not be carried as-is into the new main-based LazyCat release. Closing it does NOT mean transient retry support is already implemented upstream or is inherently unnecessary: current delivery still has no bounded retry loop. Any future retry work should be a separately tested minimal change against the new user/device consumer, preserving revalidation, event/device deduplication and delivery semantics.

No branch deletion and no changes to App notification-click PR #106 or wrapper flock fallback.

PR #3118, opened 2026-09-20
fix(coding-agent): refresh context after native compact

wtj-0527:
## Summary

- propagate Codex native compaction `afterTokens` into Studio session context usage
- refresh the client context meter when a native `/compact` command completes
- consume Codex native `token_count` events as the authoritative active-context value
- persist active context separately from cumulative/request billing usage
- avoid reporting a single post-compact token snapshot as both Before and After
- clear stale in-memory working state after synchronous native compaction completes

## Root cause

Codex successfully wrote a native `compacted` checkpoint, but Studio did not copy the resulting context token count into durable session state. Worse, normal run settlement rebuilt `contextTokens` from the latest proxy billing row (`input + cached input + output`). That value is request/cumulative accounting, not the native thread's active context, so the UI could jump from a real post-compact 26k snapshot back to 500k+ and stay wrong after restart.

## Fix

- add nullable `sessions.context_tokens` for the active native/window estimate
- update it only from authoritative context events (`/compact` result or Codex native token-count frames)
- keep proxy/session billing totals limited to `inputTokens` and `outputTokens`
- restore `contextTokens` from the dedicated persisted field, never from the latest billing row

## Verification

- `npm test -- --run tests/server/run-chat-load-state.test.ts tests/server/agent-runner-utils.test.ts tests/server/run-chat-usage.test.ts tests/server/coding-agent-session-command.test.ts tests/server/codex-compact.test.ts` — 97 passed
- `npm run build` — passed
- `git diff --check` — passed


wtj-0527:
Closed without merge per user decision. The context-meter/context-recovery changes are no longer being pursued because hiding the meter when native context was unknown caused an unacceptable UI regression. No further fix or packaging is planned from this PR.

PR #3133, opened 2026-09-20
fix(settings): default new session push on and retain saved preferences

wtj-0527:
New sessions default push enabled unless explicitly false/0, including SQLite and fallback storage. Existing stored 0 values are deliberately preserved because historical default-off cannot be distinguished from explicit opt-out. UI new-session creation no longer turns missing preference into false. Existing global config remains under persistent appHome/config.json with locked merge and backups; added reload/unrelated-patch regression preserving false settings. Focused 4 tests passed. No mass migration or deployment. Hermes external history without a local settings row is unchanged.


wtj-0527:
已rebase main 002fcec16217d50459721381147f9bb2658e2ecc，并修复原CI唯一失败：hermes-schemas.test.ts:182仍断言新增push_enabled列默认0，现与默认开启契约对齐为1。已有显式0保持不变的测试保留。Focused schema/store/global-config 10/10通过。注意本机全量coverage并未全绿（33文件/69失败，含client mock与其他测试）；未据此声称全量通过，等待本次远端CI重新执行。未部署、冻结LPK/App不变。

